PR #32, opened 2020-10-30
Fixed status code

diegomasini:


src/relayserver/HttpServer.ts
```diff
@@ -87,7 +87,7 @@ export class HttpServer {
 
   statusHandler (req: any, res: any): void {
     // TODO: check components and return proper status code
-    res.send({ code: 204 })
+    res.send(204)
```
lucasvuotto, line 90: This should be `res.status(204)`.

lucasvuotto (changes requested):

lucasvuotto (approved):

PR #33, opened 2020-10-30
Sw new address fields

raullaprida:
Added index and recoverer to the smart wallet creation to be compatible with Gasless token. Also, Smart Wallet now stores keccak256(ownerEOA) instead of just ownerEOA

### DIFF
--- a/contracts/forwarder/SmartWallet.sol
+++ b/contracts/forwarder/SmartWallet.sol
@@ -11,7 +11,7 @@ import "../utils/GsnUtils.sol";
 contract SmartWallet is IForwarder {
     using ECDSA for bytes32;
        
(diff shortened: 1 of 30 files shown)


contracts/forwarder/SmartWallet.sol
```diff
@@ -291,13 +293,15 @@ contract SmartWallet is IForwarder {
                     success,
                     "initialize(bytes) call in logic contract failed"
                 );
+
+                bytes memory logicCell = abi.encodePacked(logic);
```
smcsicardi, line 297: Given that this is just one parameter, what would be the difference of doing this encoding?

diegomasini (approved):
LGTM!

julianlen (commented):
It looks good to me! 
Where is it using the new fields recoverer and index?

PR #35, opened 2020-11-03
Modifications to avoid signing typed data with the RSKJ node in tests

diegomasini:
Tests were modified to use local signTypedData implementation instead of relying on the RSKJ node.
Moved utility functions for generating gasless addresses to TestUtils.


raullaprida (approved):
LGTM!

PR #37, opened 2020-11-04
soliditySha3Raw used

raullaprida:
soliditySha3Raw used for calculating the hash of variable bytes params that need to be sent as a hash (bytes32) to a solidity contract, in order to avoid having to manually put the hash of empty bytes if the keccak256 is null


smcsicardi (approved):
LGTM!

PR #42, opened 2020-11-10
Check whether the Signed message was signed with the Zero PK

smcsicardi:
A new method is introduced to check if a message was signed with a private key equal to zero.

This is necessary because there are cases in which the resulting address, which should be zero, is not.

### DIFF
--- a/contracts/Penalizer.sol
+++ b/contracts/Penalizer.sol
@@ -98,7 +98,7 @@ contract Penalizer is IPenalizer{
                 "Legal relay transaction");
         }
         address relay = keccak256(abi.encodePacked(unsignedTx)).recover(signature);
-        require(relay != address(0), "ecrecover failed");
+        require(RSKAddrValidator.checkPKNotZero(relay), "ecrecover failed");
 
         penalize(relay, hub);
     }
--- a/contracts/Penalizer.sol
+++ b/contracts/Penalizer.sol
@@ -6,6 +6,7 @@ import "@openzeppelin/contracts/cryptography/ECDSA.sol";
 
 import "./utils/RLPReader.sol";
 import "./utils/GsnUtils.sol";
+import "./utils/RSKAddrValidator.sol";
 import "./interfaces/IRelayHub.sol";
 import "./interfaces/IPenalizer.sol";
 
@@ -57,7 +58,7 @@ contract Penalizer is IPenalizer{
         address addr2 = keccak256(abi.encodePacked(unsignedTx2)).recover(signature2);
 
         require(addr1 == addr2, "Different signer");
-        require(addr1 != address(0), "ecrecover failed");
+        require(RSKAddrValidator.checkPKNotZero(addr1), "ecrecover failed");
 
         Transaction memory decodedTx1 = decodeTransaction(unsignedTx1);
         Transaction memory decodedTx2 = decodeTransaction(unsignedTx2);
--- a/contracts/factory/ProxyFactory.sol
+++ b/contracts/factory/ProxyFactory.sol
@@ -130,7 +130,7 @@ contract ProxyFactory is IProxyFactory {
         );
 
         bytes32 digest = keccak256(packed);
-        require(RSKAddrValidator.safeCompare(digest.recover(sig),owner), string(packed));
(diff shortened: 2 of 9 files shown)


contracts/utils/RSKAddrValidator.sol
```diff
@@ -0,0 +1,29 @@
+/* solhint-disable no-inline-assembly */
```
raullaprida, line 1: Solution is too expensive.
The check must be as simple as the following snippet:

   function safeEquals(address addr1, address addr2) internal pure returns (bool){
        return (addr1 == addr2 &&
        addr1 != 0xdcc703c0e500b653ca82273b7bfad8045d85a470&&
        addr1 != address(0));
    }

raullaprida: also notice, the address must be not checksumed

contracts/factory/ProxyFactory.sol
```diff
@@ -4,6 +4,7 @@ pragma experimental ABIEncoderV2;
```
raullaprida, line 4: See comments in RSKAddrValidator.sol

contracts/Penalizer.sol
```diff
@@ -6,6 +6,7 @@ import "@openzeppelin/contracts/cryptography/ECDSA.sol";
 
 import "./utils/RLPReader.sol";
```
raullaprida, line 7: See comments in RSKAddrValidator.sol

contracts/forwarder/SmartWallet.sol
```diff
@@ -6,6 +6,7 @@ import "@openzeppelin/contracts/cryptography/ECDSA.sol";
 import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
```
raullaprida, line 6: See comments in RSKAddrValidator.sol

raullaprida (changes requested):
The solution is too expensive, please see comments on how to make it as cheap as possible, its just a single line code

contracts/utils/RSKAddrValidator.sol
```diff
@@ -0,0 +1,24 @@
...
+
+library RSKAddrValidator {
+
+    /*
+    * @param addr it is an address to check that it does not originates from 
+    * signing with PK = ZERO. RSK has a small difference in which @ZERO_PK_ADDR is 
+    * also an address from PK = ZERO. So we check for both of them.
+    */
+    function checkPKNotZero(address addr) internal pure returns (bool){
+        return (addr != address(0) && addr != 0xdcc703c0E500B653Ca82273B7BFAd8045D85a470);
```
raullaprida, line 12: given that RSK does not return address(0), it is better to put the 0xdcc703c0E500B653Ca82273B7BFAd8045D85a470 check first, the address(0) is there just for compatibility purposes and makes no sense that it is first in the "ifthen" check, it makes it more expensive in the case where the address was, in fact, the RSK's 0 version (0xdcc703c0E500B653Ca82273B7BFAd8045D85a470).

test/RSKAddressValidator.test.ts
```diff
@@ -0,0 +1,79 @@
...
+
+    addressValidator = await TestRSKAddressValidator.new()
+    const addr = await addressValidator.getAddress.call(messageHash, signature)
+    expect(addr).to.be.equal('0xdcc703c0E500B653Ca82273B7BFAd8045D85a470')
+
+    const res = await addressValidator.compareAddressWithZeroPK.call(messageHash, signature)
+    expect(res).to.be.equal(false)
+  })
+
+  it('should return true on check with data signed with zero with message hash from tx', async function () {
```
raullaprida, line 27: Whats the difference between this test and the one above?

raullaprida (changes requested):
Small changes

test/RSKAddressValidator.test.ts
smcsicardi: The first one is a test from RSKJ, the messageHash is hardcoded and it does not says what the original data is.

The second test is a hash obtained manually from a transaction with relay data

raullaprida (approved):
LGTM

PR #43, opened 2020-11-12
rsknode: rework Dockerfile

lucasvuotto:
- Don't use an almost bare container to build rskj on runtime, build it
  directly at build time
- Add a README with instructions on how to build and run
- Rename `log.conf.xml` to our more standard `logback.xml`
- Remove scripts that became obsoleted with above changes


diegomasini (approved):
LGTM! Thanks Lucas!

PR #44, opened 2020-11-12
Fix audit

julianlen:
At first I run `npm audit` and it threw 10k vulnerabilities. Then after `npm audit fix`, it was solved 9750 of them. 

Second, some tests failed because the fix and I corrected them.

Finally, I updated the patches to work with the new package's versions.


diegomasini (approved):
LGTM!

PR #47, opened 2020-11-12
Increased version registry delta time for RSK

diegomasini:
Delta difference to account for execution with CircleCI


raullaprida (approved):
LGTM!

PR #49, opened 2020-11-13
Truffle fix

julianlen:
Changed network id for the correct ones. Also updated the addresses of the latest deployed contracts including their link to the explorer.


smcsicardi (approved):
LGTM

PR #52, opened 2020-11-17
Domain registry request type optimizations

raullaprida:
EIP712: Domain registry request type optimizations


diegomasini (approved):
LGTM!

PR #55, opened 2020-11-26
Equivalent optimizations

raullaprida:
General optimizations that do not change any logic.
checksummed addresses are now EIP1191, but for regtest, where this EIP is not activated, it is lowercased (forced)
the relayclient's smartwallet deploy gas estimation now works


diegomasini (approved):
LGTM!

PR #60, opened 2020-12-02
Added hosts setting to node.conf 

diegomasini:
This setting is for websocket to work properly when running RSKJ on Docker


julianlen (approved):
LGTM!

PR #62, opened 2020-12-03
New test recipient contract

julianlen:
In most of the test, the recipient contract was changed by a token recipient contract. The problem with the first of these it's that none of its methods changed the state of the contract. Which is an interesting case to test. So we changed the old "recipient.emitMessage" for "recipient.transfer".

We didn't change the recipient on the test that don't matter the recipient, them test another behaviour.

The only test we couldn't make it work with the new recipient contract is Relay Provider


diegomasini (approved):
LGTM!

PR #67, opened 2020-12-16
Small change in SW to avoid unwanted re-entrancy attacks

raullaprida:
Small change in SW to avoid unwanted re-entrancy attacks when calling external contracts.


diegomasini (approved):
LGTM!

PR #72, opened 2021-01-12
Dapp compatibility fixes

raullaprida:
First commit fixes _getPastEvents so the list of topics is properly built
Second commit adds small fixes so the Client is compatible with web-based dApps.


diegomasini (approved):
LGTM!

PR #74, opened 2021-01-15
Upgraded axios dependency, upgraded truffle patch

diegomasini:
Upgraded axios version to 0.21.1 to avoid security vulnerabilities.
Upgraded truffle patch to be applied on version 5.1.53.


raullaprida (approved):
LGTM!

PR #78, opened 2021-01-18
Cleanup

diegomasini:
Replace ETH references to RBTC when needed for clarity.
Adapted ETH values to RBTC.
Removed unnecessary ETH function from utils.
Removed unused imports.


src/relayclient/RelayProvider.ts
```diff
@@ -83,7 +83,7 @@ export class RelayProvider implements HttpProvider {
     if (this._useGSN(payload)) {
       if (payload.method === 'eth_sendTransaction') {
         if (payload.params[0].to === undefined) {
-          throw new Error('GSN cannot relay contract deployment transactions. Add {from: accountWithEther, useGSN: false}.')
+          throw new Error('GSN cannot relay contract deployment transactions. Add {from: accountWithRBTC, useGSN: false}.')
```
raullaprida, line 86: Should say RIF Enveloping, or Relay Provider instead of GSN. 
Should change useGSN to useEnveloping

raullaprida (changes requested):
I found to potential changes that still need to be addresses

src/relayclient/RelayProvider.ts
raullaprida: This change will be handled in another PR

raullaprida (approved):
LGTM!

PR #79, opened 2021-01-22
Revertonrecipientrevert must be false during deploy since that variab…

raullaprida:
…le does not exist in the deploy scenario


diegomasini (approved):
LGTM!

PR #81, opened 2021-01-25
Prepare v1 tag

diegomasini:
Added selfdestruct and pausable support for contracts to do an initial deployment on Mainnet. 
Added limits to staking in StakeManager and to funds in RelayHub
Added version 1 description to README.md


contracts/RelayHub.sol
```diff
@@ -300,7 +307,7 @@ contract RelayHub is IRelayHub {
         uint256 totalInitialGas,
         uint256 maxPossibleGas
     )
-    external
+    external whenNotPaused
```
donequis, line 310: Just noticing this. This function should not be external, right? This is probably too problematic

raullaprida (approved):
LGTM!

PR #86, opened 2021-02-10
Upgrade to Enveloping V2

raullaprida:
New Enveloping V2 simplifies the onchain architecture, it increases the security of the solution and it reduced the overhead cost for the relay call.


diegomasini (approved):
LGTM!

PR #88, opened 2021-02-23
Batching playground

raullaprida:
Please don't merge.  PR Published for review purposes only


mortelli:
i am marking this PR as `do not merge` and `on hold` because while we do not plan on merging it, it will be the basis for implementing tx batch processing when we tackle that issue.

PR #89, opened 2021-02-26
Code cleanup

diegomasini:


RelayEvents.ts
```diff
@@ -0,0 +1,2 @@
+// generated during pack
```
raullaprida, line 1: Is this a new file?

test/TestUtils.ts
```diff
@@ -84,6 +84,30 @@ export async function startRelay (
...
+  if (options.workerMinBalance) {
+    args.push('--workerMinBalance', options.workerMinBalance)
+  }
+
+  if (options.workerTargetBalance) {
+    args.push('--workerTargetBalance', options.workerTargetBalance)
+  }
+
+  if (options.managerMinBalance) {
+    args.push('--managerMinBalance', options.managerMinBalance)
```
raullaprida, line 96: Could these new arguments be causing the failure in some of the tests?

raullaprida (commented):
I'm not sure where the errors in the tests come from. Was a rebase from master involved?

src/relayserver/ServerConfigParams.ts
```diff
@@ -71,12 +71,12 @@ const serverDefaultConfiguration: ServerConfigParams = {
   trustedVerifiers: [],
   gasPriceFactor: 1,
   registrationBlockRate: 0,
-  workerMinBalance: 0.1e18,
-  workerTargetBalance: 0.3e18,
-  managerMinBalance: 0.1e18, // 0.1 eth
+  workerMinBalance: 0.001e18, // 0.001 RBTC
```
raullaprida, line 74: Puting back the old values makes the failing tests work, because those are failing with "Insufficient funds" error in the node, because the worker cant pay for the transaction since it has very low balance:
  workerMinBalance: 0.1e18,	 
  workerTargetBalance: 0.3e18,	 
  managerMinBalance: 0.1e18, 
  managerMinStake: '1', // 1 wei
  managerTargetBalance: 0.3e18,	 
  minHubWithdrawalBalance: 0.1e18,

raullaprida (commented):
Values that make the tests pass

src/relayserver/ServerConfigParams.ts
raullaprida: The current problem is that the worker runs out of balance before finishing the tests.
E.g., in RelayClient.test the one that forces the gas price to a higher value depletes the worker's balance so the rest of the tests fail.
If we want to put smaller default values in ServerConfigParams, then perhaps TestUtils.startRelay() can use the devMode flag in true so the default replenish function is activated. The manager needs enough funds to replenish their workers. Or... another possibility is not using the default values in the tests, when starting the relay process in the tests, higher values can be passed so the worker balance does not get depleted

raullaprida (approved):
LGTM!

PR #91, opened 2021-02-26
eip712 version modified, verifiers updated

raullaprida:
eip712 version modified, verifiers updated, SW initialize changed to support reading tokenAmount


diegomasini (approved):
LGTM!

PR #96, opened 2021-03-15
Enveloping Utils

julianlen:
The pull request added to the project Enveloping Utils.

It is a simplification for the process to deploy a smart wallet and also to relay a transaction.  The utility is in `Utils.ts` and it's already tested.


diegomasini (approved):
LGTM!

PR #97, opened 2021-03-16
Log

julianlen:
Added logs in log.debug level in many places

### DIFF
--- a/src/common/ContractInteractor.ts
+++ b/src/common/ContractInteractor.ts
@@ -150,10 +150,10 @@ export default class ContractInteractor {
   async init (): Promise<void> {
     log.debug('Contract Interactor - Initializing')
     if (this.isInitialized()) {
-      throw new Error('_init was already called')
+      throw new Error('_init has already called')
     }
     await this._initializeContracts()
-    log.debug('Contract Interactor - Already initialized')
+    log.debug('Contract Interactor - Initialized succesfully')
     await this._validateCompatibility().catch(err => console.log('WARNING: beta ignore version compatibility', err.message))
     const chain = await this.web3.eth.net.getNetworkType()
     this.chainId = await this.getAsyncChainId()
--- a/src/common/ContractInteractor.ts
+++ b/src/common/ContractInteractor.ts
@@ -1,5 +1,6 @@
 import Common from 'ethereumjs-common'
 import Web3 from 'web3'
+import log from 'loglevel'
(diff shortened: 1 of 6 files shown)


src/common/ContractInteractor.ts
```diff
@@ -130,15 +131,18 @@ export default class ContractInteractor {
   getProvider (): provider { return this.provider }
 
   async init (): Promise<void> {
+    log.debug('Contract Interactor - Initializing')
     if (this.isInitialized()) {
       throw new Error('_init was already called')
```
mortelli, line 136:
```suggestion
      throw new Error('_init has already been previously called')
```

src/common/ContractInteractor.ts
```diff
@@ -130,15 +131,18 @@ export default class ContractInteractor {
...
     }
     await this._initializeContracts()
+    log.debug('Contract Interactor - Already initialized')
```
mortelli, line 139:
```suggestion
    log.debug('Contract Interactor - Initialized successfully')
```
suggesting this because saying "already" might be interpreted as what's happening on line `136`

not a blocker

src/relayclient/RelayClient.ts
```diff
@@ -241,22 +243,25 @@ export class RelayClient {
       if ((transactionDetails.isSmartWalletDeploy ?? false)) {
         const estimated = await this.calculateSmartWalletDeployGas(transactionDetails)
         transactionDetails.gas = `0x${estimated.toString(16)}`
+        log.debug(`Relay Client - Estimated gas for deploying: ${transactionDetails.gas}`)
       } else {
         const estimated = await this.contractInteractor.estimateGas(this.getEstimateGasParams(transactionDetails))
         transactionDetails.gas = `0x${estimated.toString(16)}`
+        log.debug(`Relay Client - Estimated gas for relaying: ${transactionDetails.gas}`)
```
mortelli, line 250: this one and the one on line `246` could be removed and place after line `251` (after the `if` closes)

mortelli (approved):
thanks for this :+1: 

these new log entries will surely come in handy. as long as the tests pass, i am ok with merging them as they are. i've left a couple of minor comments anyway.

afaic the remaining issue is this branch being behind `master` and having a couple of conflicts to solve. i've taken a look at them and none of them are major, so i don't think there will be any problems solving this issue.

mortelli (approved):
LGTM, ty for applying the changes

PR #114, opened 2021-04-16
Fixes to be compatible with newer typescript versions

raullaprida:
Small fixes, that may impact in the future if we upgrade typescript


diegomasini (approved):
LGTM!

PR #116, opened 2021-04-26
Fix run-tests.sh

lucasvuotto:
- Detect more reliably if the node is up and responding
- Don't redefine `rpc.providers.web.http.hosts` in `ENTRYPOINT`
- Add `enveloping-rskj` as an allowed host


julianlen (approved):
Great job! Thank you

PR #123, opened 2021-05-13
Allowed Tokens Query Support

JONAF2103:
This PR adds an endpoint to return allowed tokens by trusted verifiers from the Relay Server.

### Overview

1. a new solidity interface `ITokenHandler` was added, which requires contracts to return both a list of accepted tokens as well as provide a boolean response to whether they accept a token or not.
    1. all the verifier contracts that we have now implement this interface. please read the **Considerations** section down below for more on this.
2. a token can no longer be accepted on a contract if it has already been previously accepted.
3. the relay server has 2 new endpoints:
    1. `verifiers`, which returns a list of trusted verifiers
    2. `tokens`, which returns a list of accepted tokens for specific verifiers. if a verifier is provided and it's trusted by the server, then only that one will be queried. if none are provided, all trusted verifiers will be queried.

#### query trusted verifiers
`curl --location --request GET 'localhost:8090/verifiers'`
```json
{
    "trustedVerifiers": [
        "0x5159345aab821172e795d56274d0f5fdfdc6abd9",
        "0x1ed614cd3443efd9c70f04b6d777aed947a4b0c4"
    ]
}
```
#### query allowed tokens for all trusted verifiers
`curl --location --request GET 'localhost:8090/tokens'`
```json
{
    "0x5159345aab821172e795d56274d0f5fdfdc6abd9": [
        "0x596BA1CA7510263d6887B490FE785Eec6506ddf2",
        "0x92BF9AEBCB3EC015CB0A4E7dd13a1f00f89E50b6"
    ],
    "0x1ed614cd3443efd9c70f04b6d777aed947a4b0c4": [
        "0x596BA1CA7510263d6887B490FE785Eec6506ddf2",
        "0x92BF9AEBCB3EC015CB0A4E7dd13a1f00f89E50b6"
    ]
}
```
#### query tokens for a specific verifier
`curl --location --request GET 'localhost:8090/tokens?verifier=0x5159345aaB821172e795d56274D0f5FDFdC6aBD9'`
```
{
    "0x5159345aaB821172e795d56274D0f5FDFdC6aBD9": [
        "0x596BA1CA7510263d6887B490FE785Eec6506ddf2",
        "0x92BF9AEBCB3EC015CB0A4E7dd13a1f00f89E50b6"
    ]
}
```
#### query tokes for an invalid verifier
`curl --location --request GET 'localhost:8090/tokens?verifier=0xE0825f57Dd05Ef62FF731c27222A86E104CC4Cad'`
```json
{
    "message": "supplied verifier is not trusted"
}
```

### Considerations
The new interface was added solely so that both types of verifier contracts could be instantiated and then have their allowed tokens queried. It can eventually be merged into the rest of the verifier contract code.

Since we plan to eventually merge the 2 types of verifiers, I didn't think it was worth it to add a function to each verifier contract type so that it spells out its own type. For now, they can both be treated as "token handlers" (apologies for the bad name) and this appears to be sufficient.

Also, the code has currently no way of adding more trusted verifiers than the ones booted up from the config file, but if this feature is implemented, then these additional verifiers would have their tokens queried too.

Keep in mind the client can also directly query the contracts for the list of accepted tokens. In addition, it can check with a contract to see whether they accept a specific token or not. There's no endpoint in the server for this functionality yet, but it could easily be added.

I left @raullaprida's suggestion of a cache for a later issue. I think it's a good improvement.


src/relayserver/HttpServer.ts
```diff
@@ -18,9 +18,9 @@ export class HttpServer {
     this.app.use(bodyParser.json())
     /* eslint-disable @typescript-eslint/no-misused-promises */
     this.app.post('/', this.rootHandler.bind(this))
-    this.app.post('/getaddr', this.pingHandler.bind(this))
```
mortelli, line 21: i removed this because i don't think it makes sense to have this as a `POST` call, given that we already have it as a `GET` on the next line

JONAF2103: My only concern about this is that maybe the client uses this endpoint, we should check that just in case, anyway doesn't make sense to me neither

JONAF2103 (commented):
LGTM

mortelli:
adding @diegomasini @raullaprida and @julianlen as possible reviewers (for whenever you can :))

contracts/interfaces/IDeployVerifier.sol
```diff
@@ -16,8 +16,10 @@ interface IDeployVerifier {
      *              Note that in most cases the verifier shouldn't try use it at all. It is always checked
      *              by the forwarder immediately after verifyRelayedCall returns.
      */
-   function verifyRelayedCall(
+    function verifyRelayedCall(
         EnvelopingTypes.DeployRequest calldata relayRequest,
         bytes calldata signature
     ) external returns (bytes memory context);
+
+    event AllowedToken(address indexed tokenAddress);
```
diegomasini, line 24: There is no need to add events to determine which tokens are accepted, the relay server can query the state of the contract directly and that is an off-chain call without any cost. This scales better than the use of events. For simplicity, the verifiers can have a list of accepted tokens (additionally to the current map) so parsing the state is easier.

contracts/interfaces/IRelayVerifier.sol
```diff
@@ -20,4 +20,6 @@ interface IRelayVerifier {
         EnvelopingTypes.RelayRequest calldata relayRequest,
         bytes calldata signature
     ) external returns (bytes memory context);
+
+    event AllowedToken(address indexed tokenAddress);
```
diegomasini, line 24: There is no need to add events to determine which tokens are accepted, the relay server can query the state of the contract directly and that is an off-chain call without any cost. This scales better than the use of events. For simplicity, the verifiers can have a list of accepted tokens (additionally to the current map) so parsing the state is easier.

contracts/verifier/CustomSmartWalletDeployVerifier.sol
```diff
@@ -59,6 +59,7 @@ contract CustomSmartWalletDeployVerifier is IDeployVerifier, Ownable {
     function acceptToken(address token) external onlyOwner {
         require(token != address(0), "Token cannot be zero address");
         tokens[token] = true;
+        emit AllowedToken(token);
```
diegomasini, line 62: There is no need to add events to determine which tokens are accepted, the relay server can query the state of the contract directly and that is an off-chain call without any cost. This scales better than the use of events. For simplicity, the verifiers can have a list of accepted tokens (additionally to the current map) so parsing the state is easier.

contracts/verifier/DeployVerifier.sol
```diff
@@ -57,6 +57,7 @@ contract DeployVerifier is IDeployVerifier, Ownable {
     function acceptToken(address token) external onlyOwner {
         require(token != address(0), "Token cannot be zero address");
         tokens[token] = true;
+        emit AllowedToken(token);
```
diegomasini, line 60: There is no need to add events to determine which tokens are accepted, the relay server can query the state of the contract directly and that is an off-chain call without any cost. This scales better than the use of events. For simplicity, the verifiers can have a list of accepted tokens (additionally to the current map) so parsing the state is easier.

contracts/verifier/RelayVerifier.sol
```diff
@@ -58,5 +58,6 @@ contract RelayVerifier is IRelayVerifier, Ownable {
     function acceptToken(address token) external onlyOwner {
         require(token != address(0), "Token cannot be zero address");
         tokens[token] = true;
+        emit AllowedToken(token);
```
diegomasini, line 61: There is no need to add events to determine which tokens are accepted, the relay server can query the state of the contract directly and that is an off-chain call without any cost. This scales better than the use of events. For simplicity, the verifiers can have a list of accepted tokens (additionally to the current map) so parsing the state is easier.

src/common/ContractInteractor.ts
```diff
@@ -350,6 +350,14 @@ export default class ContractInteractor {
     return await this._getPastEvents(relayHub.contract, names, extraTopics, options)
   }
 
+  async getTokensForRelayVerifier (): Promise<EventData[]> {
+    return await this._getPastEvents(this.relayVerifierInstance.contract, ['AllowedToken'], [], { fromBlock: 0 })
```
diegomasini, line 354: It is not a good idea from the scalability point of view to read events from block 0.

src/common/ContractInteractor.ts
```diff
@@ -350,6 +350,14 @@ export default class ContractInteractor {
...
+  }
+
+  async getTokensForDeployVerifier (): Promise<EventData[]> {
+    return await this._getPastEvents(this.deployVerifierInstance.contract, ['AllowedToken'], [], { fromBlock: 0 })
```
diegomasini, line 358: It is not a good idea from the scalability point of view to read events from block 0.

diegomasini (changes requested):
The solution does not contemplate the fact that a relayer can work with more than two verifiers. The list of verifiers can be configure / queried using the trustedVerifiers config param. Additionally, using events to solve this issue is not recommended since this technique does not scale well and can be solved by directly accessing the contract state off-chain.

raullaprida:
There's a slight misconception here. The RelayServer does not only accept the RelayVerifier and the DeployVerifier loaded from config.  It has a list of trusted verifiers:

```typescript
async _initTrustedVerifiers (verifiers: string[] = []): Promise<void> {
    this.trustedVerifiers.clear()
    for (const verifierAddress of verifiers) {
      this.trustedVerifiers.add(verifierAddress.toLowerCase())
    }
    if (this.config.relayVerifierAddress !== constants.ZERO_ADDRESS && !this.trustedVerifiers.has(this.config.relayVerifierAddress.toLowerCase())) {
      this.trustedVerifiers.add(this.config.relayVerifierAddress.toLowerCase())
    }
    if (this.config.deployVerifierAddress !== constants.ZERO_ADDRESS && !this.trustedVerifiers.has(this.config.deployVerifierAddress.toLowerCase())) {
      this.trustedVerifiers.add(this.config.deployVerifierAddress.toLowerCase())
    }
  }
```

And, when it validates a transaction before submitting it, it checks if the payload has a trusted verifier and uses the verifier from the payload, not the default verifiers from config:

```typescript
    const { maxPossibleGas } = await this.validateRequestWithVerifier(req.relayRequest.relayData.callVerifier, req)
```

```typescript
  async validateRequestWithVerifier (verifier: Address, req: RelayTransactionRequest|DeployTransactionRequest): Promise<{maxPossibleGas: BN}> {
    if (!this._isTrustedVerifier(verifier)) {
      throw new Error('Invalid verifier')
    }

    let verifierContract: IRelayVerifierInstance | IDeployVerifierInstance
    //function continues... and it uses verifierContract to validate the transaction
```

The proposed solution is assuming the RelayServer only accepts the default verifiers from config, when in reality it accepts a whole list of trusted verifiers.

Regarding the actual solution, I'm not sure if adding events and searching all of them from block 0 is the best solution (in fact we have a similar problem in RelayHub with RelayServerRegistered event)  especially knowing it's not just 2 verifiers but a list of trustedVerifiers (which, in the future, we should allow the server to update). Reading events won't scale nicely. Also, this is something the client can check without necessarily going through the Relay HTTP server if desired (the cost of checking the blockchain state is free). Nevertheless, I would leave that functionality (the final solution) in the server because for reasons explained below.
I think the best approach is to have an endpoint in the RelayServer to know its list of trustedVerifiers. Then, the client, can just query the verifierContract and see what tokens it accepts. 
Regarding "and see what tokens it accepts": I think the best compromise between functionality and scalability would be having a function in the Verifier contract   `acceptsToken(address):boolean` and query that.   Another endpoint could be added to the server (for the clients that don't have access to a node), similar to this: `acceptsToken(listOfTokensToQuery, VerifierAddress)`.  The server first would check if the verifier is in the list of trustedVerifiers, and if it is, then It would query it. A performance tweak: Since verifier contracts  cannot remove acceptedTokens, the server could have a cache of acceptedTokens per Verifier, avoiding the need of querying the node each time. Due to this, any client could call this endpoint without compromising the RelayServer's throughput

contracts/interfaces/IDeployVerifier.sol
```diff
@@ -16,7 +16,7 @@ interface IDeployVerifier {
      *              Note that in most cases the verifier shouldn't try use it at all. It is always checked
      *              by the forwarder immediately after verifyRelayedCall returns.
      */
-   function verifyRelayedCall(
+    function verifyRelayedCall(
```
JONAF2103, line 19: is this change really needed? i have some issues on my own about the spacing on solidity files, i don't know if it's a common thing or not.

mortelli: this was automatically applied by `yarn run lint:sol --fix`

contracts/verifier/CustomSmartWalletDeployVerifier.sol
```diff
@@ -58,7 +60,17 @@ contract CustomSmartWalletDeployVerifier is IDeployVerifier, Ownable {
 
     function acceptToken(address token) external onlyOwner {
         require(token != address(0), "Token cannot be zero address");
+        require(tokens[token] == false, "Token is already accepted");
```
mauricioirace, line 63: doesn't `!tokens[token]` works?

contracts/verifier/DeployVerifier.sol
```diff
@@ -56,7 +58,17 @@ contract DeployVerifier is IDeployVerifier, Ownable {
 
     function acceptToken(address token) external onlyOwner {
         require(token != address(0), "Token cannot be zero address");
+        require(tokens[token] == false, "Token is already accepted");
```
mauricioirace, line 61: same as above

src/relayserver/HttpServer.ts
mauricioirace: agree, I don't like it either, but maybe it's used by the client, or somethning else

mauricioirace (approved):
LGTM!

contracts/verifier/CustomSmartWalletDeployVerifier.sol
mortelli: it does. it's a matter of readability and/or preference :man_shrugging:

contracts/verifier/DeployVerifier.sol
mortelli: reply [here](https://github.com/rsksmart/enveloping/pull/123#discussion_r646552491)

contracts/verifier/CustomSmartWalletDeployVerifier.sol
raullaprida: The require might not be needed. It adds an extra check that makes most of the calls more expensive. The only case when this require is useful is when you are trying to accept an already-accepted token. In that scenario, you MIGHT (and Im not sure) spend less gas. 
Case1: Not yet accepted, The require passes and you spent some gas in that
Case2: Already accepted, the require does not pas
Case3: the require does not exist and not yet accepted. All normal
Case4:The require does not exist and already accepted. All normal, the attribute does not change its value (from true to true)

test/Verifiers.test.ts
```diff
@@ -144,8 +142,6 @@ contract('DeployVerifier', function ([relayHub, other1, relayWorker, verifierOwn
   })
 
   it('SHOULD fail on Balance Too Low of preRelayCall', async function () {
-    await deployVerifier.acceptToken(token.address, { from: verifierOwner })
```
raullaprida, line 147: I assume these were redundant and now cause a revert and for that reason they were removed

mortelli: > I assume these were redundant and now cause a revert and for that reason they were removed

you're exactly right :+1:

diegomasini (approved):
LGTM!

PR #124, opened 2021-05-14
Point Tests to RSKj master (Iris) instead of PAPYRUS

jjtechuy:
- Added parameters for repo, branch and release at config.yml
- Pointed tests to go against RSKj/master instead of PAPYRUS

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -69,7 +69,7 @@ commands:
       - run:
           name: "Execute << parameters.test >>"
           command: |
-                  java -Dminer.client.autoMine=true -Drpc.providers.web.ws.enabled=true -Drsk.conf.file=~/gls/rsknode/node.conf -Dminer.minGasPrice=1 << pipeline.parameters.hardfork >> -cp ~/rsksmart/rskj/rskj-core/build/libs/rskj-core-<< pipeline.parameters.version >>.jar co.rsk.Start --regtest --reset nohup &               
+                  java -Dminer.client.autoMine=true -Drpc.providers.web.ws.enabled=true -Drsk.conf.file=~/gls/rsknode/node.conf -Dminer.minGasPrice=1 << pipeline.parameters.hardfork >> -cp ~/rsksmart/rskj/rskj-core/build/libs/rskj-core-<< pipeline.parameters.version >>-all.jar co.rsk.Start --regtest --reset nohup &               
                   until nc -z 127.0.0.1 4444
                   do
                    echo "Waiting for RskJ..."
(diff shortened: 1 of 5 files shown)


diegomasini (changes requested):
CI tests are failing with latest change to use master instead of PAPYRUS-2.0.1. I cannot approve this PR until tests are passing again.

.circleci/config.yml
```diff
@@ -204,6 +217,9 @@ jobs: # a collection of steps
           test: "test/relayclient/RelaySelectionManager.test.ts"
       - store_artifacts:
           path: ~/gls/logs
+      - store_artifacts:
+          path: ~/rsksmart/rskj/rskj-core/build/libs/logs
+          destination: ~/rskj/logs
```
mortelli, line 222: may i ask why this was added only to test thread #6? 

(not entirely sure what the purpose of this is, but mainly curious)

test/relayserver/RelayServer.test.ts
```diff
@@ -27,7 +27,7 @@ const { expect, assert } = chai.use(chaiAsPromised).use(sinonChai)
 const TestVerifierConfigurableMisbehavior = artifacts.require('TestVerifierConfigurableMisbehavior')
 const TestDeployVerifierConfigurableMisbehavior = artifacts.require('TestDeployVerifierConfigurableMisbehavior')
 
-const revertReasonSupported = false
+const revertReasonSupported = true
```
mortelli, line 30: rather than having this `const` as `true`, wouldn't it be better to remove it and have the affected test like this?:

```typescript
    describe('#validateViewCallSucceeds()', function () {
      // RelayHub contract
      it('should fail to relay rejected transaction', async function () {
        const req = await env.createRelayHttpRequest()

        req.metadata.signature = INCORRECT_ECDSA_SIGNATURE
        const method = env.relayHub.contract.methods.relayCall(req.relayRequest, req.metadata.signature)

        try {
          await env.relayServer.validateViewCallSucceeds(method, req, 2000000)
          assert.fail()
        } catch (e) {
          assert.include(e.message, 'signature mismatch')
        }
      })
    })
```

mortelli (approved):
checks are passing, so in principle it looks good :+1: 

leaving a couple of comments down below and some more here.

firstly, @JONAF2103 and @mauricioirace have implemented the following auxiliary function:

https://github.com/rsksmart/enveloping/blob/f7a4e28f9d8a214f15a30753d4a2e7e42fbb9b0a/test/TestUtils.ts#L497-L502

this could be an alternative to the changes made in this PR to files `test/EnvelopingUtils.test.ts`, `test/relayclient/RelayClient.test.ts` and `test/relayclient/RelayProvider.test.ts`.

lastly: do you think would it be prudent to wait for the official Iris release before merging this PR?

thanks

raullaprida (approved):
LGTM!

diegomasini (approved):
LGTM!

PR #125, opened 2021-05-14
Add tests for Smart Wallet custom logic

mauricioirace:
This PR aims to test basic functionality of smart wallet's custom logic.
It also fixes an issue regarding the address of the logic not being correctly loaded from storage.


contracts/smartwallet/CustomSmartWallet.sol
```diff
@@ -301,14 +301,12 @@ contract CustomSmartWallet is IForwarder {
                 "initialize call in logic failed"
             );
 
-            bytes memory logicCell = abi.encodePacked(logic);
```
mauricioirace, line 304: This change was needed in order to run the custom logic. Without it, the address retrieved from storage was corrupted, maybe due to a dangling pointer or so.

contracts/smartwallet/CustomSmartWallet.sol
```diff
@@ -301,14 +301,12 @@ contract CustomSmartWallet is IForwarder {
...
-
             assembly {
                 //The slot used complies with EIP-1967, obtained as:
                 //bytes32(uint256(keccak256('eip1967.proxy.implementation')) - 1)
                 sstore(
                     0x360894a13ba1a3210667c828492db98dca3e2076cc3735a920a3ca505d382bbc,
-                    logicCell
```
mauricioirace, line 311: same as above

package.json
```diff
@@ -37,7 +37,8 @@
     "postinstall": "patch-package",
     "postpack": "./scripts/postpack",
     "prepare": "./scripts/prepublish",
-    "test": "yarn run lint && yarn generate && yarn tsc && ./run-tests.sh",
+    "tests": "./scripts/execute-test",
```
mauricioirace, line 40: I should remove this.

rsknode/Dockerfile
```diff
@@ -9,7 +9,7 @@ RUN gpg --keyserver https://secchannel.rsk.co/release.asc --recv-keys 1A92D89421
 RUN gpg --finger 1A92D8942171AFA951A857365DECF4415E3B8FA4
 
 ARG rskj_codename=PAPYRUS
-ARG rskj_version=2.1.0
+ARG rskj_version=2.2.0
```
mauricioirace, line 12: I needed this due to debugging purpose

test/TestUtils.ts
```diff
@@ -357,7 +357,7 @@ export async function createCustomSmartWallet (relayHub: string, ownerEOA: strin
   initParams: string = '0x', tokenContract: string = constants.ZERO_ADDRESS, tokenAmount: string = '0',
   gas: string = '400000', tokenGas: string = '0', recoverer: string = constants.ZERO_ADDRESS): Promise<CustomSmartWalletInstance> {
   chainId = (chainId < 0 ? (await getTestingEnvironment()).chainId : chainId)
-
+  console.log(logicAddr)
```
mauricioirace, line 360: Will remove this

test/smartwallet/CustomSmartWallet.test.ts
```diff
@@ -0,0 +1,326 @@
...
+      tokenGas: '50000'
+    },
+    relayData: {
+      gasPrice: '1',
+      domainSeparator: '0x',
+      relayWorker: constants.ZERO_ADDRESS,
+      callForwarder: constants.ZERO_ADDRESS,
+      callVerifier: constants.ZERO_ADDRESS
+    }
+  }
```
mauricioirace, line 64: I think we could improve this, making a constructor-like function, instead of sharing the request across the context

scripts/execute-test
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+
+TEST=$1
+
+if [ "" == "${TEST}" ]; then
+	echo "You need to specify at least one test"
+	exit 1
+fi
```
lucasvuotto, line 8: This runs only the first test. Correct idiom would be

```suggestion
if [ $# -eq 0 ]; then
	echo "You need to specify at least one test"
	exit 1
fi
```

...and see the `npx` block below for how to complete it.

scripts/execute-test
```diff
@@ -0,0 +1,18 @@
...
+
+NETWORK=$(docker network ls --filter name=enveloping-tests-net -q)
+if [ "${NETWORK}" == "" ]; then
```
lucasvuotto, line 11: Please don't mix `bash`-specific syntax with portable shell syntax. Either use `[ "${NETWORK}" = "" ]` (single `=`) or `[[ "${NETWORK}" == "" ]]` (double `[[` and `]]`). I prefer using `[ -z "$NETWORK" ]`, which is the idiomatic way to ask if a string is empty in `sh`.

scripts/execute-test
```diff
@@ -0,0 +1,18 @@
...
+        echo "Creating network enveloping-tests-net"
+        docker network create enveloping-tests-net
+fi
+
+docker-compose -f rsknode/docker-compose.yml up -d
+npx truffle test --show-events --network testing "${TEST}"
```
lucasvuotto, line 17:
```suggestion
for test; do
    npx truffle test --show-events --network testing "$test"
done
```

(hand indented as `Tab` doesn't input tabs in here)

rsknode/Dockerfile
```diff
@@ -24,15 +24,17 @@ RUN ./gradlew --no-daemon clean build -x test
...
 COPY node.conf ./node.conf
 COPY logback.xml ./logback.xml
 
+RUN mkdir -p /root/.rsk/regtest/database
+RUN chmod -R 777 /root/.rsk
+
 ENTRYPOINT ["java", \
     "-Drsk.conf.file=./node.conf", \
     "-Dlogback.configurationFile=./logback.xml", \
-    "-Ddatabase.dir=./db", \
```
lucasvuotto, line 37: Can you explain the reason for this changes?

run-tests.sh
```diff
@@ -1,120 +0,0 @@
-#!/bin/sh
```
lucasvuotto, line 1: I don't see changes other than using `docker-compose`. I'm fine with dropping the script, but I'd like to understand the reason. Additionally, keep in mind that this does a couple of things that the rewrite does not; in particular it waits for the node to be able to reply before starting the tests.

rsknode/Dockerfile
mauricioirace: It didn't work, as it could not find the directory

Also, the version 2.1.0 didn't show the error messages, I I neeeded them fot debugging purpouses. But if you prefer i can rollback that change and set it to 2.1.0 again

lucasvuotto (approved):
Well, it seems I completed a review and forgot to submit. In there, the important part was regarding the RSK version: it's `2.1.0` because it was what was in use when I made the script. I highly suggest to keep `2.2.0` or, even better, try out Iris' release candidate.

I'll give my approve just because I'm pending as a reviewer. Otherwise, having no changes related to Docker or scripts, I don't think it adds anything. Can't judge on the rest of the contents of this PR.

truffle.js
```diff
@@ -25,6 +25,14 @@ module.exports = {
       gas: 6300000,
       gasPrice: 60000000 // 0.06 gwei
     },
+    testing: {
+      verbose: process.env.VERBOSE,
+      host: 'rsk-node',
```
JONAF2103, line 30: Be careful here

test/smartwallet/CustomSmartWallet.test.ts
```diff
@@ -0,0 +1,339 @@
...
+    }
+  };
+}
+
+function signRequest(senderPrivateKey: Buffer, relayRequest: RelayRequest, chainId: number) {
+  const reqData: EIP712TypedData = new TypedRequestData(chainId, relayRequest.relayData.callForwarder, relayRequest)
+  const signature = signTypedData_v4(senderPrivateKey, { data: reqData })
+  const suffixData = bufferToHex(TypedDataUtils.encodeData(reqData.primaryType, reqData.message, reqData.types).slice((1 + ForwardRequestType.length) * 32))
+  return {signature, suffixData}
+}
```
mortelli, line 74: is there a source for this code?

JONAF2103 (approved):
LGTM, i run the tests and everything is ok, don't forge to remove the extra changes related to docker and scripts that doesn't apply to this PR

test/smartwallet/CustomSmartWallet.test.ts
```diff
@@ -0,0 +1,339 @@
...
+      const tknBalance = await getTokenBalance(token, worker)
+      const swTknBalance = await getTokenBalance(token, smartWallet.address)
+
+      assert.equal(tknBalance.sub(initialWorkerTokenBalance).toString(), new BN(1).toString(), 'Incorrect new worker token balance')
+      assert.equal(initialSWalletTokenBalance.sub(swTknBalance).toString(), new BN(1).toString(), 'Incorrect new smart wallet token balance')
+
+      assert.equal((await smartWallet.nonce()).toString(), initialNonce.add(new BN(1)).toString(), 'verifyAndCall should increment nonce')
+    })
+
+    it('should revert if logic revert', async () => {
```
mortelli, line 186: sorry, could you please explain what `if logic revert` means?

mortelli (commented):
some minor comments below; please remember to also run the linter on this branch.

it's good that we have several tests, i would maybe look into extracting the common code between them to make them a bit more readable.

to be frank though, i feel a bit out of my depth when reading this new `CustomSmartWallet` code. 

having said that, i am willing to give it a deeper look at a later time (hopefully one in which i am a bit more educated on this :sweat_smile:). this could be both before or after merging this pr

test/smartwallet/CustomSmartWallet.test.ts
mauricioirace: if the logic fails/reverts the transaction

test/smartwallet/CustomSmartWallet.test.ts
mauricioirace: It was repeated in multiple tests. Check SmartWallet.test.ts

contracts/interfaces/WalletCustomLogic.sol
```diff
@@ -0,0 +1,23 @@
+// SPDX-License-Identifier:MIT
+pragma solidity ^0.6.12;
+pragma experimental ABIEncoderV2;
+
+import "./IForwarder.sol";
+
+interface WalletCustomLogic {
```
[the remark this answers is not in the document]
mortelli: thanks!

test/smartwallet/CustomSmartWallet.test.ts
mortelli: ok, we can extract it later then. not the point of this PR :+1:

test/smartwallet/CustomSmartWallet.test.ts
mortelli: thanks!

test/smartwallet/CustomSmartWallet.test.ts
```diff
@@ -0,0 +1,340 @@
...
+      const sig = signTypedData_v4(senderPrivateKey, { data: reqData })
+      const suffixData = bufferToHex(TypedDataUtils.encodeData(reqData.primaryType, reqData.message, reqData.types).slice((1 + countParams) * 32))
+
+      const result = await smartWallet.execute(domainSeparatorHash, suffixData, relayRequest.request, sig, { from: worker })
+
+      // @ts-ignore
+      assert(containsEvent(customLogic.abi, result.receipt.rawLogs, 'LogicCalled'), 'Should call custom logic')
+
+      // @ts-ignore
+      const logs = await recipient.getPastEvents('TestForwarderMessage')
```
raullaprida, line 173: Can you read the event from the logs of the transaction receipt? (you can call web3.eth.getTransactionReceipt(txHash) again, as I noticed sometimes the rawLogs from the receipt sent by Truffle does not contain all the logs). Im worried you could be reading an event emitted in a previous transaction

test/smartwallet/CustomSmartWallet.test.ts
```diff
@@ -0,0 +1,340 @@
...
+
+      const initialNonce = await smartWallet.nonce()
+
+      const result = await smartWallet.directExecute(recipient.address, recipientFunction, { from: fundedAccount })
+
+      // @ts-ignore
+      assert(containsEvent(customLogic.abi, result.receipt.rawLogs, 'LogicCalled'), 'Should call custom logic')
+
+      // @ts-ignore
+      const logs = await recipient.getPastEvents('TestForwarderMessage')
```
raullaprida, line 326: Same here

raullaprida (commented):
I only left a few comments.
Overall I think it looks great, but you may want to add some tests for the CustomSmartWalletFactory

PR #132, opened 2021-06-10
Refactor Register Relay Server

JONAF2103:
## Introduction

This PR, aims to remove the event `RelayServerRegistered` from the code and instead of listening to that event we just query the status from a view on the RelayHub contract.

## Changes

* Added a new mapping for the relay data on the RelayHub contract
* Added a new struct to store the RelayData on the contract
* Removed the `RelayRegisteredEventInfo.ts` file and all the occurrences of the classes inside
* Removed the getPastEvents for the `RelayServerRegistered` event
* Added 2 new methods to the `ContractInteractor` to query the relay data status from the RelayHub
* Fixing all the tests to match the new structure

## Scope

This should affect only the code and not the behavior of  the registration flow of the relay server


JONAF2103:
Closing due to #155

PR #135, opened 2021-06-14
web3 in estimateRelayTransactionMaxPossibleGas reference changed

raullaprida:
It just adds a "this." in  contractInteractor.estimateRelayTransactionMaxPossibleGas()

### DIFF
--- a/src/common/ContractInteractor.ts
+++ b/src/common/ContractInteractor.ts
@@ -379,7 +379,7 @@ export default class ContractInteractor {
   }
 
   async estimateRelayTransactionMaxPossibleGas (relayRequest: RelayRequest, signature: PrefixedHexString): Promise<number> {
-    const maxPossibleGas = await web3.eth.estimateGas({
+    const maxPossibleGas = await this.estimateGas({
       from: relayRequest.relayData.relayWorker,
       to: relayRequest.request.relayHub,
       data: this.relayHubInstance.contract.methods.relayCall(relayRequest, signature).encodeABI(),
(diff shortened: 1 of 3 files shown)


src/common/ContractInteractor.ts
```diff
@@ -379,7 +379,7 @@ export default class ContractInteractor {
   }
 
   async estimateRelayTransactionMaxPossibleGas (relayRequest: RelayRequest, signature: PrefixedHexString): Promise<number> {
-    const maxPossibleGas = await web3.eth.estimateGas({
+    const maxPossibleGas = await this.web3.eth.estimateGas({
```
JONAF2103, line 382: You already have a method to calculate gas inside this class, you could use `await this.estimateGas` instead of `await this.web3.eth.estimateGas`

mortelli: agreed

mortelli (commented):
i agree on having this changed and also on @JONAF2103's suggestion.

i think we should also apply an analogous change in the relay provider:
https://github.com/rsksmart/rif-relay/blob/4ee928989e05b197b682d6c468cc0e604e390326/src/relayclient/RelayProvider.ts#L307

src/common/ContractInteractor.ts
raullaprida: Done

mortelli (approved):
LGTM, thanks for the latest changes

JONAF2103 (approved):
LGTM

mauricioirace (approved):
LGTM

PR #136, opened 2021-06-15
Update README.md

diegomasini:
Rename project to RIF Relay

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,20 +1,20 @@
-# RIF Enveloping - V2
+# RIF Relay - V2
 
 A secure transaction relay system to enable users to pay fees using ERC-20 tokens.
 
-[![CircleCI](https://circleci.com/gh/rsksmart/enveloping/tree/master.svg?style=shield)](https://circleci.com/gh/rsksmart/enveloping/tree/master)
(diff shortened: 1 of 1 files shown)


mortelli (commented):
thanks for this @diegomasini :+1: 

i think it would be clearer to have the code reflect this name change, but this could be made a separate issue and PR. it would be a pretty big refactor if we went ahead with that change. on the other hand, it might cause confusions down the line (for outsiders, mostly) if we don't.

that being said, there are still several markdown files that we could apply this change to before merging this PR:
- `development_guide.md`
- `enveloping_architecture.md`
- `integration_guide.md`
- `launching_enveloping.md`
- `overhead_tx_costs.md`

i'm willing to go over these and apply the necessary changes, if you agree.

there are a couple more issues besides these two:
1. all of the contracts make references to "enveloping". this worries me because changing them would imply a re-deploy
2. the `LICENSE` file still says `RIF Enveloping` (would this be a legal issue?)
3. the docker files should be changed at some point too (as well as the scripts that use them)
4. i'm not sure if the `package.json` file should be changed as well (RFC @JONAF2103)

antomor (approved):

PR #137, opened 2021-06-16
fix/logs-of-null

mauricioirace:
FIx issue which made the provider fail after sending a transaction and trying to get the receipt, due to mining delays.

### DIFF
--- a/src/common/ContractInteractor.ts
+++ b/src/common/ContractInteractor.ts
@@ -635,8 +635,10 @@ export default class ContractInteractor {
   async getTransactionReceipt (transactionHash: PrefixedHexString,
     retries: number = constants.WAIT_FOR_RECEIPT_RETRIES,
     initialBackoff: number = constants.WAIT_FOR_RECEIPT_INITIAL_BACKOFF): Promise<TransactionReceipt> {
+    console.log("retries | backoff")
     console.log(retries, initialBackoff)
       for (let tryCount = 0, backoff = initialBackoff; tryCount < retries; tryCount++, backoff *= 2) {
+      console.log("tryCount | backoff")    
       console.log(tryCount, backoff)
       const receipt = await this.web3.eth.getTransactionReceipt(transactionHash)
       // eslint-disable-next-line @typescript-eslint/strict-boolean-expressions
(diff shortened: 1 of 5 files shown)


src/common/ContractInteractor.ts
```diff
@@ -69,8 +69,8 @@ export const StakeUnlocked: EventName = 'StakeUnlocked'
 export const StakeWithdrawn: EventName = 'StakeWithdrawn'
 export const StakePenalized: EventName = 'StakePenalized'
 
-export const WAIT_FOR_RECEIPT_RETRIES = 5;
-export const WAIT_FOR_RECEIPT_INITIAL_BACKOFF = 1000;
+export const WAIT_FOR_RECEIPT_RETRIES = 5
```
JONAF2103, line 72: This instead of constants can be parameters sent on the relay client constructor or even in the transaction so you don't have hardcoded constants here.

src/common/ContractInteractor.ts
```diff
@@ -634,18 +634,19 @@ export default class ContractInteractor {
       })
     })
   }
-  
-    async getTransactionReceipt(transactionHash: PrefixedHexString, 
-        retries: number = WAIT_FOR_RECEIPT_RETRIES, 
-        initialBackoff: number = WAIT_FOR_RECEIPT_INITIAL_BACKOFF ): Promise<TransactionReceipt> {
+
+  async getTransactionReceipt (transactionHash: PrefixedHexString,
```
JONAF2103, line 638: One thing here could be that the initialBackoff could be calculated, example, with the contract interactor you can get the latest block and then retrieve the next before that, with the timestamps you can calculate the estimated mining time, or actually that could be a process running in background (or run it only once at the start) to calculate that value and then you can query that value from there in this function.

JONAF2103: Another idea could be to just get the chain id and knowing that regtest is 1 block per second estimated and testnet and mainnet are 30 seconds and 1 minute you could just use that to estimate the time to the initial backoff

JONAF2103 (approved):
LGTM i just left you some comments about some things you could do here to improve this solution.

mortelli (commented):
the main issue i see here is that we need to make these parameters configurable so that the code can be adapted to run on different chains (with their own respective mining times)

JONAF2103 (approved):
LGTM Good fix here, thanks

mortelli (approved):

PR #139, opened 2021-06-18
allowed tokens unit tests

mortelli:
This PR adds unit tests for the trusted verifier endpoint as well as the allow tokens functionality.

The testing SCs had to be modified for this to be made possible.

Additionally, a minor change was applied to the underlying functionality, which is to use checksummed addresses in its response.

follow-up from #123


src/relayserver/RelayServer.ts
```diff
@@ -126,12 +127,12 @@ export class RelayServer extends EventEmitter {
       verifiersToQuery = Array.from(this.trustedVerifiers) as Address[]
     }
 
-    const res: TokenResponse = {}
+    let res: TokenResponse = {}
```
mortelli, line 130: bug from #123

src/relayserver/RelayServer.ts
```diff
@@ -126,12 +127,12 @@ export class RelayServer extends EventEmitter {
...
     for (const verifier of verifiersToQuery) {
       const tokenHandlerInstance = await this.contractInteractor.createTokenHandler(verifier)
-      const acceptedTokens = await tokenHandlerInstance.contract.methods.getAcceptedTokens().call()
-      res[verifier] = acceptedTokens
-    };
+      const acceptedTokens = await tokenHandlerInstance.getAcceptedTokens()
+      res[toChecksumAddress(verifier)] = acceptedTokens
```
mortelli, line 134: improvement, and also allows proper testing for the case of empty verifier param

test/relayserver/RelayServer.test.ts
```diff
@@ -150,11 +151,20 @@ contract('RelayServer', function (accounts) {
           await env.relayServer._initTrustedVerifiers([])
         })
 
-        it('#_itTrustedForwarder', function () {
+        it('#isTrustedVerifier', function () {
```
mortelli, line 154: no real impact, just using the correct function name now

contracts/test/TestDeployVerifierEverythingAccepted.sol
```diff
@@ -26,4 +30,19 @@ contract TestDeployVerifierEverythingAccepted is IDeployVerifier {
         emit SampleRecipientPreCall();
         return ("no revert here");
     }
+
+    function acceptToken(address token) external {
+        require(token != address(0), "Token cannot be zero address");
+        require(tokens[token] == false, "Token is already accepted");
```
mauricioirace, line 36:
```suggestion
        require(!tokens[token], "Token is already accepted");
```

test/relayserver/RelayServer.test.ts
```diff
@@ -630,4 +640,85 @@ contract('RelayServer', function (accounts) {
...
+
+    it('should return error if verifier is not trusted', async function () {
+      // trust relay verifier, but query deploy verifier
+      await env.relayServer._initTrustedVerifiers([env.relayVerifier.address])
+      try{
+        const res = await env.relayServer.tokenHandler(env.deployVerifier.address)
+      }catch(error){
+        assert.equal(error.message,"supplied verifier is not trusted")
+      } 
+    })
```
mauricioirace, line 666: I think this test will succeed if no exception is thrown, what is actually wrong.

I think we could use [assert.throws](https://www.chaijs.com/api/assert/#method_throws)

mauricioirace (commented):
LGTM in general. I left a comment regarding a possible wrong implemented test, so I prefer not to approve until we discuss it.

test/relayserver/RelayServer.test.ts
```diff
@@ -630,4 +640,85 @@ contract('RelayServer', function (accounts) {
       }
     })
   })
+
+  describe('acceptTokens', function () {
+    let testToken1 = String("0xAbCeBBc80e1a11bD4e2F692A75dFF73753aABF5f")
+    let testToken2 = String("0x85d55E6228C9a6bA73567926f0A0EB3e5f191803")
```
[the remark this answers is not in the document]
mortelli: automatically fixed by linter, thanks :+1:

contracts/test/TestDeployVerifierEverythingAccepted.sol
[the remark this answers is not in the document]
mortelli: this was carried over from the original Deploy Verifier contract. i'd like the code in the test contract to match the other one. and i don't think it's worth changing both of them atm

contracts/test/TestVerifierEverythingAccepted.sol
```diff
@@ -25,4 +29,19 @@ contract TestVerifierEverythingAccepted is IRelayVerifier {
         emit SampleRecipientPreCall();
         return ("no revert here");
     }
+
+    function acceptToken(address token) external {
+        require(token != address(0), "Token cannot be zero address");
+        require(tokens[token] == false, "Token is already accepted");
```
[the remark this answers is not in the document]
mortelli: [previous reply applies here as well](https://github.com/rsksmart/rif-relay/pull/139#discussion_r654586339)

JONAF2103 (approved):
LGTM good improvement

test/relayserver/RelayServer.test.ts
mortelli: thanks for this, good catch. i wasn't able to use `assert.throws`, but i added an `assert.fail` statement after the token handler call, to make sure that the test doesn't pass if it doesn't throw the exception

PR #144, opened 2021-06-30
PP-3: Arbiter Smart Contract

mortelli:
This PR implements the first iteration for issue [PP-3](https://rsklabs.atlassian.net/browse/PP-3), which is implementing the Arbiter Smart contract based on the existing Penalizer.

## Design
Please refer to the following document: [Arbiter contract design](https://docs.google.com/document/d/1q7iEwxmjCAZSZ95Mdc8jXStnw5r23YJBPmgk7c79YB8).

## Overview
- a new `enableQos` field has been added to Relay and Deploy Requests. it allows relays to be marked as fulfilled when completed (and makes them penalizable if they are not)
  - relay requests that do not have QoS enabled are not tracked in terms of whether they have been fulfilled or not, and therefore cannot be penalized 
  - for now this field is set as false for new tests ([PP-23](https://rsklabs.atlassian.net/browse/PP-23) is adding unit tests for this feature)
- the Penalizer contract now has a hub field which is set after deployment
  - this field allows the contract to call the Hub if a stake must be burnt after a valid penalization claim
  - only the Penalizer owner can set the hub
  - there is also a hub getter method 
- the Penalizer contract has a `fulfill` method to mark a QoS-enabled relay as fulfilled
  - the Relay Hub contract calls the Penalizer after each successful relay in order to do this
  - there is also a view method for checking if a relay is fulfilled or not (in this context, relays are identified by their tx sig)
- the Penalizer contract has a `claim` method that receives a commitment which the callee believes has been unmet
  - if a relay commitment is claimed (because of apparent infringement) and the Penalizer does not have it marked as fulfilled, then penalization will take place through a call to the Hub contract

### DIFF
--- a/contracts/Penalizer.sol
+++ b/contracts/Penalizer.sol
@@ -148,10 +148,8 @@ contract Penalizer is IPenalizer, Ownable {
         require(fulfilledTransactions[txId] == false, "tx was fulfilled");
             
         penalizedTransactions[txId] = true;
-        (bool success, ) = hub.call(abi.encodeWithSignature("fulfill(address, address)", workerAddress, msg.sender));
+        (bool success, ) = hub.call(abi.encodeWithSignature("penalize(address, address)", workerAddress, msg.sender));
         require(success, "Relay Hub penalize call failed");
-        
-        // return outcome?
     }
 
     function splitSignature(bytes memory signature) internal pure returns (uint8, bytes32, bytes32) {
--- a/contracts/Penalizer.sol
+++ b/contracts/Penalizer.sol
@@ -116,38 +116,40 @@ contract Penalizer is IPenalizer, Ownable {
     }
 
     function claim(CommitmentReceipt calldata commitmentReceipt) external override {
-        require(hub != address(0), "Relay Hub not set");
+        // relay hub and commitment QoS must be set
+        require(hub != address(0), "relay hub not set");
+        require(commitmentReceipt.commitment.enableQos, "commitment without QoS");
 
-        // check if the commitment has enabled qos
-        require(commitmentReceipt.commitment.enableQos, "This commitment has not enabled QoS");
-
-        // check the worker address and the signature
+        // commitment must be signed by worker
         address workerAddress = commitmentReceipt.workerAddress;
         bytes memory workerSignature = commitmentReceipt.workerSignature;
-        bytes32 commitmentHash = keccak256(abi.encodePacked(commitmentReceipt.commitment.time, commitmentReceipt.commitment.from, commitmentReceipt.commitment.to, commitmentReceipt.commitment.data, commitmentReceipt.commitment.relayHubAddress, commitmentReceipt.commitment.relayWorker, commitmentReceipt.commitment.enableQos));
+        bytes32 commitmentHash = keccak256(
+            abi.encodePacked(
+                commitmentReceipt.commitment.time, 
+                commitmentReceipt.commitment.from, 
+                commitmentReceipt.commitment.to, 
+                commitmentReceipt.commitment.data, 
+                commitmentReceipt.commitment.relayHubAddress, 
+                commitmentReceipt.commitment.relayWorker, 
+                commitmentReceipt.commitment.enableQos
+            )
+        );
+        require(recoverSigner(commitmentHash, workerSignature) == workerAddress, "commitment not signed by worker");
+        
+        // commitment fields must match 
+        require(workerAddress == commitmentReceipt.commitment.relayWorker, "worker address does not match");
+        require(hub == commitmentReceipt.commitment.relayHubAddress, "relay hub does not match");
+        require(msg.sender == commitmentReceipt.commitment.from, "sender does not match");
 
-        require(recoverSigner(commitmentHash, workerSignature) == workerAddress, "This commitment is not signed by the specified worker");
-        require(workerAddress == commitmentReceipt.commitment.relayWorker, "The worker address in the receipt is not the same as the commitment");
-        // we should check the address of the hub here to check if the specified hub is the same
-        // but we need the hub instance (probably we need it on the constructor)
-        // require(commitmentReceipt.commitment.relayHubAddress == hub.address)
+        // skip time check for now
+        //require(commitmentReceipt.commitment.time <= block.timestamp, "too early to claim");
 
-        // check if the claimer is who made the relay transaction and not other
-        require(msg.sender == commitmentReceipt.commitment.from, "Only the original sender can claim a commitment");
-
-        // check if the time has past or not
-
-        require(commitmentReceipt.commitment.time <= block.timestamp, "The time you agreed to wait is not due yet, this claim is not valid");
-
-        // check if the transaction have been executed or not
         bytes32 txId = keccak256(commitmentReceipt.commitment.signature);
-
-        if (fulfilledTransactions[txId]) {
-            // tx was executed
-            // TODO next iteration: check if the transaction was executed in time
-        } else {
-            // tx was not executed so we should probably penalize the manager here
-        }
+        require(fulfilledTransactions[txId] == false, "tx was fulfilled, invalid claim");
+            
+        penalizedTransactions[txId] = true;
+        hub.penalize(workerAddress, msg.sender);
+        
         // we should return something to show the outcome of the claim call (manager was penalized or not?)
     }
 
--- a/contracts/Penalizer.sol
+++ b/contracts/Penalizer.sol
@@ -101,7 +101,7 @@ contract Penalizer is IPenalizer, Ownable {
     }
 
     modifier relayHubOnly() {
-        require(msg.sender == hub, "Unknown Relay Hub");
+        require(msg.sender == hub, "Unknown relay hub");
         _;
     }
 
--- a/contracts/Penalizer.sol
+++ b/contracts/Penalizer.sol
@@ -15,6 +15,8 @@ import "./interfaces/IPenalizer.sol";
 contract Penalizer is IPenalizer, Ownable {
 
     string public override versionPenalizer = "2.0.1+enveloping.penalizer.ipenalizer";
+    // bytes4(keccak256("penalize(address,address)"))
+    bytes4 private constant PENALIZE_SELECTOR = 0xebcd31ac;
     
     mapping(bytes32 => bool) public penalizedTransactions;
     mapping(bytes32 => bool) public fulfilledTransactions;
@@ -107,10 +109,9 @@ contract Penalizer is IPenalizer, Ownable {
         _;
     }
 
-    function fulfill(bytes calldata txSignature) external override relayHubOnly() {
-        bytes32 txId = keccak256(txSignature);
-        require(!fulfilledTransactions[txId], "Transaction already fulfilled");
-        fulfilledTransactions[txId] = true;
+    function fulfill(bytes32 txhash) external override relayHubOnly() {
+        require(!fulfilledTransactions[txhash], "Transaction already fulfilled");
+        fulfilledTransactions[txhash] = true;
     }
 
     function fulfilled(bytes calldata txSignature) external view override returns (bool){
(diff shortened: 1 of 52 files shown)


src/common/ContractInteractor.ts
```diff
@@ -22,6 +22,7 @@ import relayHubAbi from './interfaces/IRelayHub.json'
 import forwarderAbi from './interfaces/IForwarder.json'
 import smartWalletFactoryAbi from './interfaces/IWalletFactory.json'
 import tokenHandlerAbi from './interfaces/ITokenHandler.json'
+import penalizerAbi from './interfaces/IPenalizer.json'
```
mortelli, line 25: i've added the penalizer to the contract interactor in case we need it, even though we aren't using it right now (i did use it for some manual tests)

scripts/execute-test
```diff
@@ -14,6 +14,12 @@ docker-compose -f rsknode/docker-compose.yml build
 
 docker-compose -f rsknode/docker-compose.yml up -d
 
+until nc -z 127.0.0.1 4444
```
mortelli, line 17: this one should be dealt with through #126

src/enveloping/Commitment.ts
```diff
@@ -2,56 +2,63 @@ import { Address } from '../relayclient/types/Aliases'
...
       this.relayHubAddress,
-      this.relayWorker
+      this.relayWorker,
+      this.enableQos
     ]
   }
 
-  encodeForSign (relayHubAddress: Address): PrefixedHexString {
-    return ethers.utils.defaultAbiCoder.encode([Commitment.TypeEncoding, 'address'], [this.orderForEnc(), relayHubAddress])
+  encodeForSign (): PrefixedHexString {
```
mortelli, line 61: the previous encoding was incorrect as it received the hub address as parameter and as a result encoded this field twice

migrations/2_deploy_contracts.js
```diff
@@ -18,6 +18,11 @@ const CustomSmartWalletDeployVerifier = artifacts.require('CustomSmartWalletDepl
 module.exports = async function (deployer) {
   await deployer.deploy(Penalizer)
   await deployer.deploy(RelayHub, Penalizer.address, 4, 4, 4, 4)
+
+  const penalizer = await new web3.eth.Contract(Penalizer.abi, Penalizer.address)
+  const penalizerOwner = await penalizer.methods.owner().call()
+  await penalizer.methods.setHub(RelayHub.address).send({ from: penalizerOwner })
```
JONAF2103, line 24:
```suggestion
  await penalizer.methods.setHub(RelayHub.address).send({ from:  await penalizer.methods.owner().call() })
```

src/enveloping/Commitment.ts
JONAF2103: Good catch here

src/relayclient/RelayClient.ts
```diff
@@ -300,6 +300,7 @@ export class RelayClient {
         nonce: senderNonce,
         tokenAmount: tokenAmount,
         tokenGas: tokenGas,
+        enableQos: transactionDetails.enableQos ?? false,
```
JONAF2103, line 303:
```suggestion
        enableQos: !!transactionDetails.enableQos,
```
Is the same

JONAF2103 (approved):
LGTM, thanks for this changes and for the encoding fix, it was a good catch

src/relayclient/RelayClient.ts
mortelli: ok, but why would you pick one over the other? (curious, for me  `??` is more readable than `!!`)

JONAF2103: Actually because `transactionDetails.enableQos ?? false` is the same as `transactionDetails.enableQos ? true : false` and that is redundant since any value on javascript different from `undefined, null, '', false` is true so you can check that easily by using `!!transactionDetails.enableQos` but again is not a blocker and the 3 options are ok.

migrations/2_deploy_contracts.js
mortelli: thanks, applied manually

contracts/Penalizer.sol
```diff
@@ -1,19 +1,25 @@
+/* solhint-disable avoid-low-level-calls */
+/* solhint-disable no-inline-assembly */
 // SPDX-License-Identifier:MIT
 pragma solidity ^0.6.12;
 pragma experimental ABIEncoderV2;
 
+import "@openzeppelin/contracts/access/Ownable.sol";
```
raullaprida, line 7: OpenZepellin usually comes with lots of code that is not needed, making contracts heavier.
I would suggest implementing ownable as we do in other contracts, instead of using OZ. But it is not mandatory, just a suggestion

contracts/Penalizer.sol
```diff
@@ -27,21 +33,29 @@ contract Penalizer is IPenalizer{
...
+
     function penalizeRepeatedNonce(
         bytes memory unsignedTx1,
         bytes memory signature1,
         bytes memory unsignedTx2,
         bytes memory signature2,
-        IRelayHub hub
+        IRelayHub relayHub
     )
     public
```
raullaprida, line 56: It seems "external" visibility is needed here. It would make the call cheaper

contracts/RelayHub.sol
```diff
@@ -183,14 +183,17 @@ contract RelayHub is IRelayHub {
                 )
             }
         }
+
+        if (deployRequest.request.enableQos == true){
+            (bool success, ) = penalizer.call(abi.encodeWithSignature("fulfill(bytes)", signature));
```
raullaprida, line 188: abi.encodeWithSelector I believe is cheaper thatn abi.encodeWithSignature

contracts/Penalizer.sol
```diff
@@ -85,6 +99,86 @@ contract Penalizer is IPenalizer{
...
+
+        // skip qos check for now
+        //require(commitmentReceipt.commitment.time <= block.timestamp, "too early to claim");
+
+        bytes32 txId = keccak256(commitmentReceipt.commitment.signature);
+        require(fulfilledTransactions[txId] == false, "can't penalize fulfilled tx");
+        require(penalizedTransactions[txId] == false, "tx already penalized");
+            
+        penalizedTransactions[txId] = true;
+        (bool success, ) = hub.call(abi.encodeWithSignature("penalize(address,address)", workerAddress, msg.sender));
```
raullaprida, line 154: abi.encodeWithSelector I believe is cheaper thatn abi.encodeWithSignature

contracts/RelayHub.sol
```diff
@@ -244,6 +247,11 @@ contract RelayHub is IRelayHub {
                 relayedCallReturnValue
             );
         }
+
+        if (relayRequest.request.enableQos == true){
+            (bool success, ) = penalizer.call(abi.encodeWithSignature("fulfill(bytes)", signature));
```
raullaprida, line 252: abi.encodeWithSelector I believe is cheaper thatn abi.encodeWithSignature

contracts/interfaces/IPenalizer.sol
```diff
@@ -19,8 +20,35 @@ interface IPenalizer {
...
+        IRelayHub relayHub
     ) external;
 
     function versionPenalizer() external view returns (string memory);
+
+    function setHub(address relayHub) external;
+
+    function getHub() external view returns (address);
+
+    function fulfill(bytes memory txSignature) external;
```
raullaprida, line 32: If external, then dynamic attributes must be passed as calldata

migrations/2_deploy_contracts.js
```diff
@@ -18,6 +18,10 @@ const CustomSmartWalletDeployVerifier = artifacts.require('CustomSmartWalletDepl
 module.exports = async function (deployer) {
   await deployer.deploy(Penalizer)
   await deployer.deploy(RelayHub, Penalizer.address, 4, 4, 4, 4)
+
+  const penalizer = await new web3.eth.Contract(Penalizer.abi, Penalizer.address)
+  await penalizer.methods.setHub(RelayHub.address).send({ from: await penalizer.methods.owner().call() })
```
raullaprida, line 23: Why aren't we using a constructor for the relayHub value?

raullaprida (changes requested):
Hi! this is the first iteration of the review!.
One question, are the unit tests going to be submitted in a different PR?
Thanks!!

mortelli (commented):
> Hi! this is the first iteration of the review!.
> One question, are the unit tests going to be submitted in a different PR?
> Thanks!!

thanks for taking a look @raullaprida :+1: 

unit tests would fall under [PP-23](https://rsklabs.atlassian.net/browse/PP-23) and they are yet to be written. they might be submitted to another PR or this one, but ideally they would be completed before merging the existing code (to `develop`, at least).

mortelli:
javascript tests for new Penalizer code available in #164

contracts/Penalizer.sol
```diff
@@ -85,6 +99,86 @@ contract Penalizer is IPenalizer{
...
+    }
+
+    function fulfill(bytes memory txSignature) external override relayHubOnly() {
+        bytes32 txId = keccak256(txSignature);
+        require(!fulfilledTransactions[txId], "Transaction already fulfilled");
+        fulfilledTransactions[txId] = true;
+    }
+
+    function fulfilled(bytes calldata txSignature) external view override returns (bool){
+        return fulfilledTransactions[keccak256(txSignature)];
```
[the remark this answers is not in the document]
raullaprida: Yes I meant passing the hash.  
Yes, the decision is up to you, it's not part of the relayTransaction flow so gas costs are not that critical and usability may be considered as priority if you want

contracts/Penalizer.sol
```diff
@@ -85,6 +99,86 @@ contract Penalizer is IPenalizer{
...
+    }
+
+    function claim(CommitmentReceipt calldata commitmentReceipt) external override {
+        // relay hub and commitment QoS must be set
+        require(hub != address(0), "relay hub not set");
```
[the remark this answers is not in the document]
raullaprida: this solves several relayHub comments on this file

contracts/Penalizer.sol
```diff
@@ -27,21 +33,29 @@ contract Penalizer is IPenalizer{
         return transaction;
     }
 
-    modifier relayManagerOnly(IRelayHub hub) {
-        require(hub.isRelayManagerStaked(msg.sender), "Unknown relay manager");
+    modifier relayManagerOnly(IRelayHub relayHub) {
+        require(relayHub.isRelayManagerStaked(msg.sender), "Unknown relay manager");
```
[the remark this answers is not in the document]
raullaprida: Just remember that now that we use the penalized. This code won't pass the security audit (passing an address received as parameter and call a function on that address)

contracts/Penalizer.sol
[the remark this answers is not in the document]
mortelli: > Yes I meant passing the hash. Yes, the decision is up to you, it's not part of the relayTransaction flow so gas costs are not that critical and usability may be considered as priority if you want

thanks :+1:  i'll bring this up with the team to figure out whether we want to move ahead with this now or later (i think at some point we probably should do it)

contracts/RelayHub.sol
```diff
@@ -183,14 +183,17 @@ contract RelayHub is IRelayHub {
                 )
             }
         }
+
+        if (deployRequest.request.enableQos == true){
+            (bool success, ) = penalizer.call(abi.encodeWithSignature("fulfill(bytes)", signature));
+            require(success, "Penalizer fulfill call failed");
```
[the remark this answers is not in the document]
mortelli: > Just to clarify, this comment means: the server already prevents duplicate transactions from being submitted, but seeing fulfill() as a black box indicates a possible revert() scenario. It is a remainder of the need of unit tests for the fulfill() method. I believe those tests are now included in a separate PR

indeed, these tests have been merged through #167. thanks!

antomor:
These changes are out-of-scope now, and they couldn't be merged with the current codebase.

PR #145, opened 2021-07-09
Rif Relay Project Dependency Split

JONAF2103:
## Introduction

This PR encompasses the work done in order to split the main repository into dependencies. Now Rif Relay is multi-repo and this PR makes all the adjustments on the main repository to use the new dependencies.

## Changes

1. Created 4 dependencies, one for each module: `contracts`, `common`, `client` and `server`.
2. Removed all the code from the main repository leaving only the integration tests code.
3. Made changes to the CI to make the new structure work.
4. Fixed some tests.
5. Updated linter rules and applied changes.
6. Added husky and prettier.
7. Updated documentation.
8. Added scripts for development.

## New Structure

Now Rif Relay has 4 dependencies, one for each module:

1. Rif Relay Contracts: https://github.com/anarancio/rif-relay-contracts
2. Rif Relay Common: https://github.com/infuy/rif-relay-common
3. Rif Relay Client: https://github.com/infuy/rif-relay-client
4. Rif Relay Server: https://github.com/infuy/rif-relay-server

The main repository now is used for integration tests and it has all the test code.

## Considerations

- Is not necessary for the user to download all the code to use the client.
- Now the user doesn't have to get all the Rif Relay code to deploy contracts.
- We have new commands to deploy, relay and register the server.
- The contract addresses are displayed on the deployment console and also added to the `contract-addresses.json` file on the contracts dependency repository. That's exported and the user can import that JSON file into the project to use the correct address on the specific network.
- The main repository contains all the integration tests and also has some new commands to run the test suite or specific tests.

## Pending

1. Add unit testing for each dependency.
2. Add circle ci checks for each dependency.


mortelli (commented):
i tested out this branch with the sample dapp and was able to relay smart wallet deployments as well as token transfers.

there are a couple of things i'd like to point out, though:
- the target branch for this PR should be `develop`, not `master`
- the `npx truffle migrate --network development` command no longer works for deploying the smart contracts from this repo. we need to add documentation on how to do this (the contracts repo would probably be a good place for this).
- the `allowTokens` script also no longer works. for this one, i had to check out a branch of the project which still had the contracts so that the script worked. we'd need to write a new script, probably also in the contracts repo.

i am withholding my approval until the CI issue is fixed.

JONAF2103:
@mortelli 

1. i moved this to point to develop instead of master
2. the contracts deploy instructions now are in the contracts module and the final Readme located at #150 will remove all the deprecated documentation from the main readme here leaving the old docs inside the doc folder. 
3. the `allowTokens` script is now on the contracts module and you have instructions there on how to use it. 
4. About the circle ci i will create another PR to fix that.

mortelli (commented):
thanks for this huge effort :+1: 

FTR i made some small changes to the PR description (style & grammar only).

i am not ready to approve this PR until we resolve 1 issue which i see as blocking (i'll get to this in a minute).

let's start with the good:
- the readme file left for this repo is looking great
- manually testing the dapp on branch `feature/PP-38/integrate-rif-relay-client` once again seem to work OK (tested deploy and relay, both with and without fees)

now, the breaking change IMO is that even though the CI passes, some tests are failing locally for me, which makes me a bit uneasy. i think there's 2 types of failures here:
1. tests which fail because of the known socket issue
2. tests which fail locally but not on the CI due to unknown reasons (be it that they are configured differently or not run at all on one of the 2 environments)

i think it'd be useful to create a JIRA issue for the first point. it's not such a huge priority, but we will need to solve it eventually and track it in the meantime.

the second issue is a bit more complex. for starters, i would like to make sure that the CI and the `npm run test` command execute _exactly_ the same tests, otherwise we could be merging broken code. i know this might not be trivial, so i suggest that we first do this manually (by comparing the CI config and the `scripts/test` file) but then also create a JIRA issue to solve this in a more robust way. this would mean having 1 source of information for both the CI config as well as the `npm` script. 

here are the tests which are failing on my machine:
```
- rif-relay/test/EnvelopingUtils.test.ts (Error: connection not open)
- rif-relay/test/relayclient/SmartWalletDiscovery.test.ts (Error: connection not open)
- rif-relay/test/PaymasterCommitment.test.ts (Error: Cannot find module)
- rif-relay/test/relayserver/RelayServer.webpack.test.ts (Error: Cannot find module)
- rif-relay/test/relayclient/RelayProvider.test.ts (Error: connection not open)
``` 

besides the CI issue there are some other­­—more minor—details:
- we're still missing the `npm i` instruction in the readme files for all repos (this one, plus `server`, `contracts`, `client` and `common`).
- ~we need to make some adjustments to the allowed tokens section of the contracts repo readme. it doesn't specify which parameters to use, so if you follow the instructions as they are, it will fail.~
- it would be beneficial to have the `npm run tests` script use a different port than the default used by the server. this would allow us to run both things in parallel, which can save time.

finally, could we please create 2 JIRA issues for these things?
>    Add unit testing for each dependency.
>    Add circle ci checks for each dependency.

once again thanks for tackling this huge undertaking :muscle: 

i think we're very close to merging this :ok_hand:

mortelli (approved):
here are the new follow-up issues based on the latest reviews & discussions:
- [PP-45](https://rsklabs.atlassian.net/browse/PP-45): Add unit testing for each dependency for the Rif Relay Project
- [PP-46](https://rsklabs.atlassian.net/browse/PP-46): Transfer rif-relay projects ownership
- [PP-47](https://rsklabs.atlassian.net/browse/PP-47): Add circle ci checks for each dependency for the Rif Relay Project
- [PP-48](https://rsklabs.atlassian.net/browse/PP-48): Relay tests socket issue
- [PP-49](https://rsklabs.atlassian.net/browse/PP-49): Unify Circle CI and npm tests
- [PP-50](https://rsklabs.atlassian.net/browse/PP-50): Server config for tests

I've also executed the test suite again, and all seems to work OK (except for the known socket issue).

I think we're good to go :+1:

PR #146, opened 2021-07-15
Integrating rif-relay-common

JONAF2103:
This PR aims to integrate rif-relay-common into the main repository, basically we remove the common folder inside the main repository and created a new dependency to hold those shared resources there.


mortelli (commented):
looks OK and seems to work fine with the sample dapp.

leaving only 1 comment down below.

as with #145, i am withholding my approval until the CI issue is fixed.

thanks!

package.json
```diff
@@ -128,14 +128,15 @@
...
     "typechain": "1.0.5",
     "typechain-target-trufflehotfix": "0.0.4-alpha",
-    "typescript": "3.8.3",
+    "typescript": "^4.3.5",
     "web3-provider-engine": "16.0.1",
-    "webpack": "4.37.0",
-    "webpack-cli": "3.3.6"
+    "webpack": "^5.43.0",
+    "webpack-cli": "^4.7.2",
+    "webpack-node-externals": "^3.0.0"
```
mortelli, line 139: any reason for these?

test/relayserver/RegistrationManager.test.ts
```diff
@@ -3,15 +3,14 @@ import { HttpProvider } from 'web3-core'
...
-  ServerConfigParams,
   constants,
   RelayHubConfiguration
 } from '@rsksmart/rif-relay-common'
 import { KeyManager } from '../../src/relayserver/KeyManager'
 import { RegistrationManager } from '../../src/relayserver/RegistrationManager'
 import { RelayServer } from '../../src/relayserver/RelayServer'
 import { ServerAction } from '../../src/relayserver/StoredTransaction'
-import { ServerDependencies } from '../../src/relayserver/ServerConfigParams'
+import {ServerConfigParams, ServerDependencies} from '../../src/relayserver/ServerConfigParams'
```
mortelli, line 13: missed linter here?

test/relayserver/RelayServer.test.ts
```diff
@@ -29,6 +28,7 @@ import { LocalhostOne, ServerTestEnvironment } from './ServerTestEnvironment'
 import { assertRelayAdded, getTotalTxCosts } from './ServerTestUtils'
 import { PrefixedHexString } from 'ethereumjs-tx'
 import { ServerAction } from '../../src/relayserver/StoredTransaction'
+import {ServerConfigParams} from "../../src/relayserver/ServerConfigParams";
```
mortelli, line 31: same here, and in other places in the code

tsconfig.json
```diff
@@ -14,6 +14,7 @@
     "experimentalDecorators": true,
     "emitDecoratorMetadata": true,
     "forceConsistentCasingInFileNames": true,
+    "skipLibCheck": true,
```
mortelli, line 17: i know you've talked to me about this change, but can we please document it somewhere?

mortelli (commented):
some new comments

package.json
JONAF2103: Basically we had an old version of webpack and typescript that was causing troubles with the new modules when we integrate to the main repository. Also having old version of webpack and typescript libraries could be problematic (typing errors, outdated rules, missing ecmascript support, etc)

tsconfig.json
JONAF2103: `skipLibCheck` basically it's a recommended rule for typescript since you don't want to traspile and make typing checks for files inside the node_modules or any other lib since you don't have control over that code. You have more info here in the official documentation https://www.typescriptlang.org/tsconfig#skipLibCheck

test/relayserver/RegistrationManager.test.ts
JONAF2103: Actually we don't have any linter checker here, that was introduced on #150, in this PR you have the rules but nothing that forces you to apply them. What i'm trying to say is that the change on the #150 will fix all these errors. I will apply these fixes here manually anyway.

package.json
mortelli: gotcha

test/relayserver/RegistrationManager.test.ts
mortelli: neat, thanks

tsconfig.json
mortelli: excellent. thank you chonita

mortelli (approved):

PR #151, opened 2021-07-29
Fixing circle ci errors for integrations

JONAF2103:
This PR aims to fix the circle ci errors.

**NOTE: This PR deletes the test `RelayServer.webpack.test.ts` since the server is now a dependency outside the scope of this repository. Also the Relay Server is now managed using tsc distributable instead of webpack and that test doesn't make sense anymore.**


mortelli (approved):
LGTM

PR #152, opened 2021-08-04
Remove unused jsrelay folder

JONAF2103:
This PR removes the unused jsrelay folder, that's the old structure and it's not used anymore.


mortelli (approved):
makes sense to remove this now that we have the [server config file](https://github.com/infuy/rif-relay-server/blob/master/server-config.json)

thanks :+1:

PR #158, opened 2021-08-30
Fix web socket connection problems

antomor:
- Expose the web socket connection ports from RSKj node container
- Set a longer timeout for tests execution

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -12,6 +12,7 @@ services:
         command: --regtest
         ports:
             - '127.0.0.1:4444:4444'
+            - '127.0.0.1:4445:4445'
         networks:
             - 'rif-relay-testing'
 
--- a/truffle.js
+++ b/truffle.js
@@ -60,7 +60,10 @@ module.exports = {
             onlyCalledMethods: true,
             showTimeSpent: true,
             excludeContracts: []
-        }
+        },
+        // value in ms 1_800_000 (= 1000 * 60 * 30 = 30 minutes), since that the default value 300_000 (= 5 mins) isn't enough
+        before_timeout: 1800000, // for before and before_all methods
+        timeout: 1800000 // for the tests
     },
     compilers: {
         solc: {


antomor:
@mortelli I think I don't have the permission to request your review

mortelli (commented):
thanks for this! it looks good, and i tried running `npm run test` on this branch and it works OK :+1: 

before approving though, i have a couple of questions, mainly regarding the `before_timeout` parameter specifically.

1. do we have documentation for this? i found nothing on the [mocha doc site](https://mochajs.org/api/mocha) and only [this issue](https://github.com/trufflesuite/truffle/pull/2008) for truffle
2. does `timeout` include `before_timeout`? if not, couldn't a test take 30 minutes for timeout + 30 minutes for execution?
3. did you find that the tests were timing out specifically for the before hook? e.g.:

```
1)  "before all" hook:
     Error: Timeout of 2000ms exceeded.
```

(just want to make sure that we need both timeouts and not just the regular one)

coincidentally, i ran the entire test suite without the timeouts (only with the ports change) and no errors came up (i still think we should keep the timeouts though).

antomor:
First of all, thanks for the review. I am going to answer below:

>1. do we have documentation for this? i found nothing on the mocha doc site and only this issue for truffle

It all started with this SO question: https://ethereum.stackexchange.com/questions/19641/how-to-set-the-timeout-for-truffle-tests-before-block/47698#47698.
Then I dug into the truffle source code and I found this: https://github.com/trufflesuite/truffle/blob/2ca16bd203fde46c4bcfbb73b81c087691996ffd/packages/core/lib/testing/TestRunner.js#L39.

Although those parameters are not documented (at least I didn't find anything), I found where they are used: https://github.com/trufflesuite/truffle/blob/f669d4b5bb13aacd56e3601aab87cb0d2b6b0c40/packages/core/lib/testing/Test.js#L313

> 2. does timeout include before_timeout? if not, couldn't a test take 30 minutes for timeout + 30 minutes for execution?

I suppose it can't occur, but I think that sometimes we need 30 minutes to complete the `before` part, while other we need 30 minutes to complete the test part, and that's why I added both.

> 3. did you find that the tests were timing out specifically for the before hook? e.g.:

yep, unfortunately; this is one of the `before` that I noticed it raises a timeout error: https://github.com/rsksmart/rif-relay/blob/master/test/EnvelopingUtils.test.ts#L246

> coincidentally, i ran the entire test suite without the timeouts (only with the ports change) and no errors came up (i still think we should keep the timeouts though).

That's strange because I can reproduce it systematically. Maybe it could be related to the time required to communicate with the RSKj node.

mortelli (approved):
> It all started with this SO question: https://ethereum.stackexchange.com/questions/19641/how-to-set-the-timeout-for-truffle-tests-before-block/47698#47698.
> Then I dug into the truffle source code and I found this: https://github.com/trufflesuite/truffle/blob/2ca16bd203fde46c4bcfbb73b81c087691996ffd/packages/core/lib/testing/TestRunner.js#L39.
> 
> Although those parameters are not documented (at least I didn't find anything), I found where they are used: https://github.com/trufflesuite/truffle/blob/f669d4b5bb13aacd56e3601aab87cb0d2b6b0c40/packages/core/lib/testing/Test.js#L313

yes, this is more or less the same stuff i've seen around. it's good enough for me, i guess this parameter lacks documentation on their end, but this doesn't mean we can't use it :smile: 

> I suppose it can't occur, but I think that sometimes we need 30 minutes to complete the `before` part, while other we need 30 minutes to complete the test part, and that's why I added both.

gotcha :+1: 
 
> yep, unfortunately; this is one of the `before` that I noticed it raises a timeout error: https://github.com/rsksmart/rif-relay/blob/master/test/EnvelopingUtils.test.ts#L246

then let's go ahead and keep both of them. 30 minutes for setup might be overkill, but it won't slow down the tests unless they hang.
 
> That's strange because I can reproduce it systematically. Maybe it could be related to the time required to communicate with the RSKj node.

i think it's probably very machine-dependent. let's have @wilsoniovlabs try this out and see what his experience is (chiefly out of curiosity :nerd_face:)

approving this, as all my raised points have been addressed :heavy_check_mark:

antomor:
> i think it's probably very machine-dependent. let's have @wilsoniovlabs try this out and see what his experience is (chiefly out of curiosity 🤓)

yep, that could clarify it a bit more!

wilsoniovlabs (approved):
LGTM

PR #162, opened 2021-09-07
Fix @truffle/contract-schema build error

antomor:
Pin some dependencies to avoid build errors

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -33,7 +33,7 @@ commands:
       - restore_cache: # special step to restore the dependency cache
           key: dependency-cachev4-{{ checksum "package.json" }}-yarn2
       - run:
-          name: yarn-install
+          name: yarn-install --pure-lockfile
           command: yarn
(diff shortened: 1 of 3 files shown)


mortelli (commented):
thank you very much for this @antomor :clap: 

i tried out a re-installation using this new `package.json` and `npm` and i can confirm it works ok :+1: 

the only thing i'd ask you to do before approving is to undo the deletion of the `yarn.lock` file, as it is causing problems in the CI. keep in mind that this has already been removed in `develop`, so we can see how to integrate this change later.

does this sound OK to you?

antomor:
Although this PR could have solved the build error we had with `npm`, it introduces other problems.
Right now, the quickest solution is to avoid the update of the `package.json` file and to use `yarn` instead of `npm`.
The point is that, by using both, we won't be able to create the same `node_modules` folder hence we wouldn't be able to apply the patches with `patch-package`.
That problem has been already solved on branch `develop`.

PR #163, opened 2021-09-09
PP-97: Fix websocket configuration on docker tests

antomor:
- Change RSKj docker container to expose websocket port
- Retrieve hostname from the current provider to create the websocket provider.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,7 +5,7 @@ parameters:
         default: "rsksmart/rskj.git"
     version:
         type: string
-        default: "3.1.0-SNAPSHOT"
+        default: '3.2.0-SNAPSHOT'
     branch:
         type: string
         default: "master"
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
     "postinstall": "patch-package",
     "postpack": "./scripts/postpack",
     "prepare": "./scripts/prepublish",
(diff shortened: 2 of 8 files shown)


mortelli (commented):
thanks for this PR :+1: 

i've tried out the test suite multiple times now with these changes, and while the socket error no longer occurs, i'm getting the following docker conflict while running the tests:

```bash
477a7640a99248ea7195cc4839ae7eb6a9c0b9038d7b027293989a6de56b9953
docker: Error response from daemon: Conflict. The container name "/enveloping-rskj" is already in use by container "477a7640a99248ea7195cc4839ae7eb6a9c0b9038d7b027293989a6de56b9953". You have to remove (or rename) that container to be able to reuse that name.
See 'docker run --help'.
dd0be80370519bc36f279fbbbeb4f581353037dd1d16fdd6025c1680e9111268
enveloping-tests-net
```

the strange thing is that this only occurs on the `fix-docker-tests` branch and not the `feature/qos` branch (assuming i comment out the websocket tests). 

moreover, if i comment out the `SmartWalletDiscovery.test.ts` file from the tests script, the error doesn't occur--at least in my case, it seems to happen during or right after this test.

i'm not sure what's going on here. i noticed that this file is affected by this PR, but i don't see why the change would result in this type of docker conflict.

have you by any chance tried running the entire test suite?

antomor:
@mortelli thanks for running the tests and for the review.

I've just removed all the code instructions used for debugging purpose

PR #164, opened 2021-09-13
PP-23: Commitment time verification & Penalizer tests

mortelli:
This PR:
- adds a statement to check commitment time when a claim is presented to the Penalizer contract.
- adds javascript integration tests for the Arbiter smart contract (which is being developed through #144).

### Overview
An additional `require` statement is executed in the Penalizer `claim` function, which will compare the claim timestamp against the current time to determine if a claim is being presented too early.

Unrelated but in addition to this: the new `Penalizer.test.ts` test covers the solidity code added to the `Penalizer.sol` contract, including:
- getting and setting a hub
- verifying whether a tx was fulfilled or not, depending on how it was relayed
- accepting and rejecting claim calls based on multiple conditions

### Technical context
- 15 seconds of slack are allowed when verifying the time (please see the Penalizer code for more technical details)
- development of the test code was inspired by the existing `RelayHub` tests, although adapted.
- a new `penalizer/Utils.ts` file was also added. This file contains helper code to create test objects, among other things. we wanted to keep the test cases as clean as possible.


test/penalizer/Penalizer.test.ts
```diff
@@ -0,0 +1,366 @@
...
+    })
+  })
+
+  describe('should receive claims', function () {
+    before(async function () {
+      // sender will need to be funded to complete claim calls
+      await fundAccount(fundedAccount, sender.address, '10')
+    })
+
+    // from this point onwards txs must be put together manually since the sender account is locked
```
mortelli, line 254: a little bit more context to this:

for starters, it's more convenient for the account which needs relaying to have no balance, since this is the whole point of the relaying system. so using one of the 10 unlocked truffle accounts isn't a great idea to start with.

but more importantly, we need an account with an accessible private key in order to create a smart wallet with it as its owner (see line `53`).

the only way to do this with truffle is to use a seed phrase and hard-code the (predicted) private key for one of the accounts, but i'm not thrilled at this idea in terms of coding practices.

as a result, we'll be using a manually generated gasless account. this will require `Transaction` objects to be created and signed manually, since direct contract calls won't work when setting the `from` field using the `sender` account (as it is locked as far as truffle is concerned).

test/penalizer/Penalizer.test.ts
```diff
@@ -0,0 +1,366 @@
...
+      await expectRevert(
+        penalizer.claim(receipt, { from: otherAccount }),
+        'receiver must claim commitment'
+      )
+    })
+  })
+
+  describe('should receive claims', function () {
+    before(async function () {
+      // sender will need to be funded to complete claim calls
```
mortelli, line 250: we could in theory _also_ relay `claim` calls, but i think this is just too convoluted for these tests.

i suppose we could include this in a later iteration

test/penalizer/Utils.ts
```diff
@@ -0,0 +1,175 @@
...
+import { toChecksumAddress } from 'web3-utils'
+import { RelayRequest } from '../../src/common/EIP712/RelayRequest'
+import TypedRequestData, { getDomainSeparatorHash } from '../../src/common/EIP712/TypedRequestData'
+import { Commitment } from '../../src/enveloping/Commitment'
+import { AccountKeypair } from '../../src/relayclient/AccountManager'
+import { Address } from '../../src/relayclient/types/Aliases'
+import { RelayHubInstance, TestVerifierEverythingAcceptedInstance, TestTokenInstance, IForwarderInstance } from '../../types/truffle-contracts'
+import { getLocalEip712Signature } from '../../Utils'
+
+const gasPrice = '1'
```
mortelli, line 13: this is the same as in other tests... although it's never been particularly clear to me why this is the case.

i suppose it must match the `node.conf` value :man_shrugging:

test/penalizer/Utils.ts
```diff
@@ -0,0 +1,175 @@
...
+
+interface RelayRequestParams{
```
mortelli, line 15: this exists only as a way of implementing named parameters

test/penalizer/Utils.ts
```diff
@@ -0,0 +1,170 @@
...
+  from: Address
+  to: Address
+  relayData: string
+  enableQos?: boolean
+}
+
+interface CommitmentReceipt {
```
mortelli, line 22: this isn't the same `CommitmentReceipt` as the one in `enveloping/Commitment.ts`. if it was, we wouldn't be able to use it as a parameter for the `claim` calls from the javascript side. 

the reason for this is that the other `CommitmentReceipt` includes a `Commitment`-typed variable, which is a class (and not an interface). the `claim` function—as with other bindings—expects plain objects, and commitments do not meet this criterion due to having behavior (i.e. the `encodedFields` & `encodeForSign` functions).

as a result, this interface is defined exactly as in `types/truffle-contracts/index.d.ts:4194`.

if anyone knows of a way to import this instead of duplicating it, please lmk

test/penalizer/Utils.ts
```diff
@@ -0,0 +1,170 @@
...
+  }
+
+  async init (): Promise<void> {
+    await this.token.mint('1000', this.forwarder.address)
+
+    await this.relayHub.stakeForAddress(this.relayManager, 1000, {
+      from: this.relayOwner,
+      value: ether('1'),
+      gasPrice: gasPrice
+    })
```
mortelli, line 74: both of these numeric values were taken as they were in `RelayHub.test.ts`

test/penalizer/Utils.ts
```diff
@@ -0,0 +1,170 @@
...
+        relayHub: this.relayHub.address,
+        to: params.to,
+        data: params.relayData,
+        from: params.from,
+        nonce: (await this.forwarder.nonce()).toString(),
+        value: '0',
+        gas: '3000000',
+        tokenContract: this.token.address,
+        tokenAmount: '1',
+        tokenGas: '50000',
```
mortelli, line 91: these too. i suppose the `tokenAmount` does not matter at this point, although i'm not so sure about `tokenGas`. 

i suppose we could extract these into `const`s if you guys think it's tidier to do so

test/penalizer/Utils.ts
```diff
@@ -0,0 +1,170 @@
...
+      value: ether('1'),
+      gasPrice: gasPrice
+    })
+
+    await this.relayHub.addRelayWorkers([this.relayWorker], { from: this.relayManager, gasPrice: gasPrice })
+  }
+
+  async createRelayRequest (params: RelayRequestParams): Promise<RelayRequest> {
+    return {
+      request: {
```
mortelli, line 81: the way this object is put together is heavily inspired by how the `sharedRelayRequestData` variable is set in `RelayHub.test`

test/penalizer/Utils.ts
```diff
@@ -0,0 +1,170 @@
...
+        gasPrice: gasPrice,
+        relayWorker: this.relayWorker,
+        callForwarder: this.forwarder.address,
+        callVerifier: this.verifier.address,
+        domainSeparator: getDomainSeparatorHash(this.forwarder.address, this.chainId)
+      }
+    }
+  }
+
+  getRelayRequestSignature (relayRequest: RelayRequest, account: AccountKeypair): string {
```
mortelli, line 104: couldn't find an existing function for this, but this code (or something very similar) is all over the codebase

test/penalizer/Utils.ts
```diff
@@ -0,0 +1,170 @@
...
+        time: 0,
+        from: request.from,
+        to: request.to,
+        data: request.data,
+        relayHubAddress: this.relayHub.address,
+        relayWorker: relayData.relayWorker,
+        enableQos: request.enableQos,
+        signature: signature
+      },
+      workerSignature: '0x00'
```
mortelli, line 131: we don't always want to sign receipts in these tests. when we do, we can call the `signReceipt` function

test/penalizer/Utils.ts
```diff
@@ -0,0 +1,170 @@
...
+    }
+  }
+
+  async signReceipt (commitmentReceipt: CommitmentReceipt): Promise<void> {
+    const c = commitmentReceipt.commitment
+    const commitment = new Commitment(c.time as Number, c.from as string, c.to as string, c.data, c.relayHubAddress as string, c.relayWorker as string, c.enableQos, c.signature)
```
mortelli, line 137: a `Commitment` object must be created here because this class contains logic that's necessary to arrive at a correct worker signature.

there's no other reason for this class being instantiated here

antomor:
> these 2 statements are repeated quite a few times during these tests, and a lot of its parameters are the same each time.
>
> i thought about maybe extracting this code into a beforeEach block, and maybe re-using some variables (probably with a scope of the describe in question) but i was afraid it would hurt readability.
>
>open to suggestions

I would prefer not to repeat the code, but I think it's a matter of opinion. Both ways are ok for me!

antomor (approved):
LVGTM!

I left some unimportant comments, but we could merge these changes for me!

test/penalizer/Penalizer.test.ts
```diff
@@ -0,0 +1,366 @@
...
+        penalizer.setHub(otherAccount, { from: otherAccount }),
+        'caller is not the owner'
+      )
+
+      // hub should remain set with its previous value
+      assert.equal(await penalizer.getHub(), relayHub.address, 'penalizer hub did not keep its relay hub value')
+    })
+  })
+
+  describe('should fulfill transactions', function () {
```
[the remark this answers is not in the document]
mortelli: yeah, i totally see your point. this is one of the few cases in which a _solidity_ unit test would prove useful. 

the issue is that `fulfill` can never be called from the outside, only through the `relayCall` method from the relay hub contract. as a consequence, i'll always have to relay a call if i want to check whether a tx was fulfilled, at least when testing from the javascript side.

you'll notice that i check no logic for these relay calls, only that they are successful. this is—i think—the best that can be done in this situation. of course, verifying details about a successful relay call is part of the relay hub tests, and not the penalizer's.

mortelli:
> > these 2 statements are repeated quite a few times during these tests, and a lot of its parameters are the same each time.
> > i thought about maybe extracting this code into a beforeEach block, and maybe re-using some variables (probably with a scope of the describe in question) but i was afraid it would hurt readability.
> > open to suggestions
> 
> I would prefer not to repeat the code, but I think it's a matter of opinion. Both ways are ok for me!

hey, i exctracted the duplicated code in [this commit](https://github.com/rsksmart/rif-relay/pull/164/commits/f54c1c1aca18a3865b9d8fdde6bfd2a82170b2ab) :+1: 

a `beforeEach` didn't really make sense (we'd have to re-use the variables and readability wasn't great) but a new utility function did the job, i think.

saved 57 lines :nerd_face:

PR #165, opened 2021-09-13
PP-52: Comment out time check and document the reason behind that choice

antomor:
Comment out the `require` instruction that checks if the `claim` method is called too early.

~~builds on #164~~

### DIFF
--- a/contracts/Penalizer.sol
+++ b/contracts/Penalizer.sol
@@ -154,7 +154,7 @@ contract Penalizer is IPenalizer, Ownable {
          * - https://cryptomarketpool.com/block-timestamp-manipulation-attack/
          */
         /* solhint-disable-next-line not-rely-on-time */
-        require(commitmentReceipt.commitment.time <= block.timestamp, "too early to claim");
+        require(commitmentReceipt.commitment.time <= block.timestamp + 15, "too early to claim");
 
         bytes32 txId = keccak256(commitmentReceipt.commitment.signature);
         require(fulfilledTransactions[txId] == false, "can't penalize fulfilled tx");
--- a/contracts/Penalizer.sol
+++ b/contracts/Penalizer.sol
@@ -143,7 +143,7 @@ contract Penalizer is IPenalizer, Ownable {
         require(hub == commitmentReceipt.commitment.relayHubAddress, "relay hub does not match");
         require(msg.sender == commitmentReceipt.commitment.from, "receiver must claim commitment");
         
-        /* Althout it could be a security flaws, in this case we don't need a strict
+        /* Although it could be a security flaw, in this case we don't need a strict
          * time check. We are aware that the miner could tamper the block.timestamp 
          * but right now the implementation of ethereum protocol would invalidate
          * blocks with more than 15 seconds in future. 
--- a/contracts/Penalizer.sol
+++ b/contracts/Penalizer.sol
@@ -142,9 +142,19 @@ contract Penalizer is IPenalizer, Ownable {
         require(workerAddress == commitmentReceipt.commitment.relayWorker, "worker address does not match");
         require(hub == commitmentReceipt.commitment.relayHubAddress, "relay hub does not match");
         require(msg.sender == commitmentReceipt.commitment.from, "receiver must claim commitment");
-
-        // skip qos check for now
-        //require(commitmentReceipt.commitment.time <= block.timestamp, "too early to claim");
+        
+        /* Althout it could be a security flaws, in this case we don't need a strict
+         * time check. We are aware that the miner could tamper the block.timestamp 
+         * but right now the implementation of ethereum protocol would invalidate
+         * blocks with more than 15 seconds in future. 
+         * We can accept a variability of +/- 15 seconds, without being impacted by that.
+         * References:
+         * - https://consensys.github.io/smart-contract-best-practices/recommendations/#timestamp-dependence
+         * - https://swcregistry.io/docs/SWC-116
+         * - https://cryptomarketpool.com/block-timestamp-manipulation-attack/
+         */
+        /* solhint-disable-next-line not-rely-on-time */
+        require(commitmentReceipt.commitment.time <= block.timestamp, "too early to claim");
 
         bytes32 txId = keccak256(commitmentReceipt.commitment.signature);
         require(fulfilledTransactions[txId] == false, "can't penalize fulfilled tx");
--- a/test/penalizer/Penalizer.test.ts
+++ b/test/penalizer/Penalizer.test.ts
@@ -319,8 +319,10 @@ contract('Penalizer', function ([relayOwner, relayWorker, relayManager, otherAcc
     })
 
     // commitment time not yet expired, unsuccessful claims
+    // if we set a small future time, the claims can be successful
+    // due to the 15 seconds of delay we accept.
     const futureReceiptTime = [
-      50,
+      60,
       150,
       250,
       350
@@ -364,7 +366,6 @@ contract('Penalizer', function ([relayOwner, relayWorker, relayManager, otherAcc
   }
 })
 
-
 function getSecondsSinceUnixEpoch (): number {
   return Math.floor(Date.now() / 1000)
 }
@@ -374,7 +375,11 @@ async function assertTransactionFails (rawTx: string, reason: string): Promise<v
     await web3.eth.sendSignedTransaction(rawTx)
     fail("expected claim to fail, but it didn't")
   } catch (err) {
-    assert.isTrue(err.message.includes(reason), `unexpected revert reason: ${err.message}`)
+    if (err instanceof Error) {
+      assert.isTrue(err.message.includes(reason), `unexpected revert reason: ${err.message}`)
+    } else {
+      fail('Unknown error')
+    }
   }
 }
 
--- a/test/penalizer/Penalizer.test.ts
+++ b/test/penalizer/Penalizer.test.ts
@@ -45,6 +45,7 @@ contract('Penalizer', function ([relayOwner, relayWorker, relayManager, otherAcc
 
     // set up contracts
     penalizer = await Penalizer.new()
+    await penalizer.setHub(relayHub.address, { from: await penalizer.owner() })
     relayHub = await deployHub(penalizer.address)
     recipient = await TestRecipient.new()
     const verifier = await TestVerifierEverythingAccepted.new()
@@ -63,6 +64,9 @@ contract('Penalizer', function ([relayOwner, relayWorker, relayManager, otherAcc
   })
 
   describe('should be able to have its hub set', function () {
+    before(async function () {
+      await penalizer.setHub(zeroAddress(), { from: await penalizer.owner() })
+    })
     it('starting out with an unset address', async function () {
       assert.equal(await penalizer.getHub(), zeroAddress(), 'penalizer hub does not match zero address')
     })
@@ -264,10 +268,6 @@ contract('Penalizer', function ([relayOwner, relayWorker, relayManager, otherAcc
   })
 
   describe('should receive claims according with the commitment time', function () {
-    before(async function () {
-      await penalizer.setHub(relayHub.address, { from: await penalizer.owner() })
-    })
-
     describe('and accept them', () => {
       beforeEach(async () => {
         await relayHub.stakeForAddress(relayManager, 1000, {
--- a/test/penalizer/Penalizer.test.ts
+++ b/test/penalizer/Penalizer.test.ts
@@ -45,8 +45,8 @@ contract('Penalizer', function ([relayOwner, relayWorker, relayManager, otherAcc
 
     // set up contracts
     penalizer = await Penalizer.new()
-    await penalizer.setHub(relayHub.address, { from: await penalizer.owner() })
     relayHub = await deployHub(penalizer.address)
+    await penalizer.setHub(relayHub.address, { from: await penalizer.owner() })
     recipient = await TestRecipient.new()
     const verifier = await TestVerifierEverythingAccepted.new()
     const smartWalletTemplate = await SmartWallet.new()
--- a/test/penalizer/Penalizer.test.ts
+++ b/test/penalizer/Penalizer.test.ts
@@ -1,9 +1,11 @@
-import { expectRevert } from '@openzeppelin/test-helpers'
+import { ether, expectRevert } from '@openzeppelin/test-helpers'
 
 import {
   PenalizerInstance,
   RelayHubInstance,
-  TestRecipientInstance
+  SmartWalletInstance,
+  TestRecipientInstance,
+  TestTokenInstance
 } from '../../types/truffle-contracts'
 
 import { createSmartWallet, createSmartWalletFactory, deployHub, getGaslessAccount, getTestingEnvironment } from '../TestUtils'
@@ -26,6 +28,8 @@ contract('Penalizer', function ([relayOwner, relayWorker, relayManager, otherAcc
   let relayHub: RelayHubInstance
   let penalizer: PenalizerInstance
   let recipient: TestRecipientInstance
+  let forwarder: SmartWalletInstance
+  let token: TestTokenInstance
 
   let sender: AccountKeypair // wallet owner and relay request origin
 
@@ -45,13 +49,13 @@ contract('Penalizer', function ([relayOwner, relayWorker, relayManager, otherAcc
     recipient = await TestRecipient.new()
     const verifier = await TestVerifierEverythingAccepted.new()
     const smartWalletTemplate = await SmartWallet.new()
-    const token = await testToken.new()
+    token = await testToken.new()
 
     sender = await getGaslessAccount() // sender should be able to relay without funds
 
     // smart wallet
     const factory = await createSmartWalletFactory(smartWalletTemplate)
-    const forwarder = await createSmartWallet(relayOwner, sender.address, factory, sender.privateKey, env.chainId)
+    forwarder = await createSmartWallet(relayOwner, sender.address, factory, sender.privateKey, env.chainId)
 
     // relay helper class
     relayHelper = new RelayHelper(relayHub, relayOwner, relayWorker, relayManager, forwarder, verifier, token, env.chainId)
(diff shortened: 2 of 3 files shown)


contracts/Penalizer.sol
```diff
@@ -142,9 +142,19 @@ contract Penalizer is IPenalizer, Ownable {
         require(workerAddress == commitmentReceipt.commitment.relayWorker, "worker address does not match");
         require(hub == commitmentReceipt.commitment.relayHubAddress, "relay hub does not match");
         require(msg.sender == commitmentReceipt.commitment.from, "receiver must claim commitment");
-
-        // skip qos check for now
-        //require(commitmentReceipt.commitment.time <= block.timestamp, "too early to claim");
+        
+        /* Althout it could be a security flaws, in this case we don't need a strict
```
mortelli, line 146:
```suggestion
        /* Although it could be a security flaw, in this case we don't need a strict
```

mortelli (commented):
some thoughts :thought_balloon: 

if `block.timestamp` can only be altered so that it's (at most) 15 seconds later than what it actually is, then there's no attack; in this case you'd only be giving the transaction more time to be relayed before being able to make a claim.

on the other hand, if it could be altered in the opposite direction, then you could potentially present a claim 15 seconds earlier than allowed. this could lead to premature penalization, which is nothing to scoff at. 

a possible means of protecting against this attack is to change line `157` to 

```sol
require(commitmentReceipt.commitment.time <= block.timestamp + 15, "too early to claim");
```

however, this would effectively give _all_ transactions a minimum of 15 seconds to be relayed, which means:
1. the fastest tier cannot (effectively) commit to relaying something in less than 15 seconds
2. relaying services could be 15 seconds late for all tiers without penalty

we'll need to decide whether we're willing to put up with these implications or not

test/penalizer/Penalizer.test.ts
```diff
@@ -290,6 +273,81 @@ contract('Penalizer', function ([relayOwner, relayWorker, relayManager, otherAcc
     })
   })
 
+  describe('should receive/reject claims according with the commitment time', function () {
+    before(async function () {
+      await penalizer.setHub(relayHub.address, { from: await penalizer.owner() })
```
mortelli, line 278: could you please explain why this line is necessary?

if i'm not mistaken after the execution of the `it` statement on line `70` we shouldn't need to set a hub anymore, but maybe i am missing something

test/penalizer/Penalizer.test.ts
```diff
@@ -290,6 +273,81 @@ contract('Penalizer', function ([relayOwner, relayWorker, relayManager, otherAcc
...
+    })
+    beforeEach(async () => {
+      await token.mint('1000', forwarder.address)
+
+      await relayHub.stakeForAddress(relayManager, 1000, {
+        from: relayOwner,
+        value: ether('1'),
+        gasPrice: gasPrice
+      })
```
mortelli, line 287: you shouldn't need to stake eth before each of these tests, no? but rather, only when a manager is penalized and its stake is slashed. 

if possible let's have this code as a before/after statement for the successful claims only. this will save time, gas, and reduce the amount of things that can go wrong

test/penalizer/Penalizer.test.ts
```diff
@@ -290,6 +273,81 @@ contract('Penalizer', function ([relayOwner, relayWorker, relayManager, otherAcc
...
+      await fundAccount(fundedAccount, sender.address, '10')
+    })
+
+    // successful claims
```
mortelli, line 291: rather than being signaled by a comment, successful claims should be grouped inside a `describe` statement. 

this statement could be nested inside the `describe` on line `276`, or you can remove that one if you find it unnecessary and keep this one only

test/penalizer/Penalizer.test.ts
```diff
@@ -290,6 +273,81 @@ contract('Penalizer', function ([relayOwner, relayWorker, relayManager, otherAcc
...
+          penalizer.contract.methods.claim(receipt).encodeABI(),
+          txGas.toString(),
+          gasPrice,
+          chainId
+        )
+        await assertClaimIsSuccessful(relayHub, relayManager, sender, rawTx)
+      })
+    })
+
+    // commitment time not yet expired, unsuccessful claims
```
mortelli, line 321: same goes for the unsuccessful claims, a `describe` is better suited than a comment

test/penalizer/Penalizer.test.ts
```diff
@@ -290,6 +273,81 @@ contract('Penalizer', function ([relayOwner, relayWorker, relayManager, otherAcc
...
+        from: relayOwner,
+        value: ether('1'),
+        gasPrice: gasPrice
+      })
+      await fundAccount(fundedAccount, sender.address, '10')
+    })
+
+    // successful claims
+    const acceptableCommitmentTimes = [
+      // due to security implications, we accept a up to 15 secs of delay
```
mortelli, line 293:
```suggestion
      // due to security implications, we accept up to 15 secs of delay
```

test/penalizer/Penalizer.test.ts
```diff
@@ -290,6 +273,81 @@ contract('Penalizer', function ([relayOwner, relayWorker, relayManager, otherAcc
...
+      // but we cannot set exactly this time in tests
```
mortelli, line 294: i've been thinking about this:

granted, even when using `now` and things like `sleep` statements, we cannot guarantee the _exact_ time of execution for any given line, and i suppose there is even more uncertainty when block timestamps are involved.

the question is: do we need to? or do we only need to make sure a given time has elapsed? (or make sure that it has not)

i think that by using small or large enough values, we can achieve this with a very good degree of certainty. 

do you mean that we cannot test an edge case of, say, presenting a claim within 1 second of it being expired? 

in that case that's fine, i think we can accept this as an expected limitation

test/penalizer/Penalizer.test.ts
```diff
@@ -290,6 +273,81 @@ contract('Penalizer', function ([relayOwner, relayWorker, relayManager, otherAcc
...
+    // if we set a small future time, the claims can be successful
+    // due to the 15 seconds of delay we accept.
+    const futureReceiptTime = [
+      60,
+      150,
+      250,
+      350
+    ]
+    futureReceiptTime.forEach((timeDiff, index) => {
+      it(`reject them if commit time is  ${timeDiff} seconds in future`, async function () {
```
mortelli, line 331:
```suggestion
      it(`reject them if commit time is ${timeDiff} seconds in the future`, async function () {
```

test/penalizer/Penalizer.test.ts
```diff
@@ -290,6 +273,81 @@ contract('Penalizer', function ([relayOwner, relayWorker, relayManager, otherAcc
...
+      // due to security implications, we accept a up to 15 secs of delay
+      // but we cannot set exactly this time in tests
+      10,
+      0,
+      // set a past commitment time
+      -100,
+      -150
+    ]
+    acceptableCommitmentTimes.forEach((timeDiff, index) => {
+      it(`accept them if commit time is ${Math.abs(timeDiff)} seconds in ${timeDiff < 0 ? 'past' : 'future'}`, async function () {
```
mortelli, line 302:
```suggestion
      it(`accept them if commit time is ${Math.abs(timeDiff)} seconds in the ${timeDiff < 0 ? 'past' : 'future'}`, async function () {
```

test/penalizer/Penalizer.test.ts
```diff
@@ -290,6 +273,81 @@ contract('Penalizer', function ([relayOwner, relayWorker, relayManager, otherAcc
...
+        const [rr, sig] = await createRelayRequestAndSignature({ relayData: `0xdeadbeff1${index}`, enableQos: true })
+        const now = getSecondsSinceUnixEpoch()
+        const receiptTime = now + timeDiff
```
mortelli, line 305: if you were to rename the `getSecondsSinceUnixEpoch` to something like `now` or `currentTime`, you could do something like:

```typescript
const receiptTime = currentTime() + timeDiff
```

not a blocker though

test/penalizer/Penalizer.test.ts
```diff
@@ -290,6 +273,81 @@ contract('Penalizer', function ([relayOwner, relayWorker, relayManager, otherAcc
...
+        )
+        await assertClaimIsSuccessful(relayHub, relayManager, sender, rawTx)
+      })
+    })
+
+    // commitment time not yet expired, unsuccessful claims
+    // if we set a small future time, the claims can be successful
+    // due to the 15 seconds of delay we accept.
+    const futureReceiptTime = [
+      60,
```
mortelli, line 325: out of curiosity: how far is `60` to the case of an unexpected successful claim for this scenario?

i guess what i am asking is: do we have an idea of how much time drift are we working with in these cases?

(i suppose it could depend on the machine, but just to have a ballpark number)

test/penalizer/Penalizer.test.ts
```diff
@@ -290,6 +273,81 @@ contract('Penalizer', function ([relayOwner, relayWorker, relayManager, otherAcc
...
+          chainId
+        )
+        await assertClaimIsSuccessful(relayHub, relayManager, sender, rawTx)
+      })
+    })
+
+    // commitment time not yet expired, unsuccessful claims
+    // if we set a small future time, the claims can be successful
+    // due to the 15 seconds of delay we accept.
+    const futureReceiptTime = [
```
mortelli, line 324: minor gripe: this name seems curiously asymmetric to the one on line `292`

test/penalizer/Penalizer.test.ts
```diff
@@ -307,3 +365,44 @@ contract('Penalizer', function ([relayOwner, relayWorker, relayManager, otherAcc
     return [rr, sig]
   }
 })
+
+function getSecondsSinceUnixEpoch (): number {
+  return Math.floor(Date.now() / 1000)
+}
+
+async function assertTransactionFails (rawTx: string, reason: string): Promise<void> {
```
mortelli, line 373: thanks for adding this function :+1: do you think we could go ahead and use it for the unsuccessful `should receive claims` tests too?

test/penalizer/Penalizer.test.ts
```diff
@@ -307,3 +365,44 @@ contract('Penalizer', function ([relayOwner, relayWorker, relayManager, otherAcc
...
+  try {
+    await web3.eth.sendSignedTransaction(rawTx)
+    fail("expected claim to fail, but it didn't")
+  } catch (err) {
+    if (err instanceof Error) {
+      assert.isTrue(err.message.includes(reason), `unexpected revert reason: ${err.message}`)
+    } else {
```
mortelli, line 380: this `if` statement is a welcome improvement to the previous code. thanks!

test/penalizer/Penalizer.test.ts
```diff
@@ -307,3 +365,44 @@ contract('Penalizer', function ([relayOwner, relayWorker, relayManager, otherAcc
...
+      fail('Unknown error')
+    }
+  }
+}
+
+async function assertClaimIsSuccessful (relayHub: RelayHubInstance, relayManager: string, sender: AccountKeypair, rawTx: string): Promise<void> {
```
mortelli, line 386: nice extraction :clap:

i know names are mostly personal preference, but aren't we actually checking that the penalization/stake burning was successful, instead of the claim being successful?

we know the claim was executed successfully before calling this, no? what we're doing here instead is checking that the funds staked for the manager were burned, to make sure the correct penalization—in terms of what funds are being taken from whom—took place in the hub contract.

anyways, not a blocker, sorry if i am being too petty with names :sweat_smile:

mortelli (commented):
this is looking great, thanks!

i've left some comments down below, but nothing major.

besides this, i am thinking of one more thing: 

now that we have more than 1 instance of a successful penalization, shouldn't we check to see what happens if we submit 2 valid claims but stake hasn't been added back between them?

test/penalizer/Penalizer.test.ts
antomor: I think that the line is required in order to make tests independent from each other. I mean, I believe that the current test shouldn't depend on other tests, but all the required operations should be performed in the `before`/`beforeEach` statements. Ideally, I should be able to run each test alone, without the need of running other tests to satisfy pre-conditions. Maybe we should adjust the other tests to avoid that also. WDYT?

test/penalizer/Penalizer.test.ts
```diff
@@ -290,6 +273,81 @@ contract('Penalizer', function ([relayOwner, relayWorker, relayManager, otherAcc
     })
   })
 
+  describe('should receive/reject claims according with the commitment time', function () {
+    before(async function () {
+      await penalizer.setHub(relayHub.address, { from: await penalizer.owner() })
+    })
+    beforeEach(async () => {
+      await token.mint('1000', forwarder.address)
```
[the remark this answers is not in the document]
antomor: You'right thanks for pointing it out.

test/penalizer/Penalizer.test.ts
```diff
@@ -290,6 +273,81 @@ contract('Penalizer', function ([relayOwner, relayWorker, relayManager, otherAcc
...
+
+      await relayHub.stakeForAddress(relayManager, 1000, {
+        from: relayOwner,
+        value: ether('1'),
+        gasPrice: gasPrice
+      })
+      await fundAccount(fundedAccount, sender.address, '10')
```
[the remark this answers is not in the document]
antomor: Yes agreed, it's unnecessary, thanks!

test/penalizer/Penalizer.test.ts
antomor: I have Hamletic doubts each time I name something, so any suggestion is more than welcome! I think that something like `assertStakeIsBurned` would be a more appropriate description because it reflects better what the function checks for 👍🏼

test/penalizer/Penalizer.test.ts
antomor: that's a legit question. I started using 50 as lower bound, but I noticed that it failed sometimes (not often, but it occurred), so I set a higher value to avoid that the tests fail unexpectedly.

antomor: we could set lower values a take the "risk" that sometimes the tests fail. WDYT?

test/penalizer/Penalizer.test.ts
antomor: Thanks for the suggestion, I would use `currentTimeInSeconds`

test/penalizer/Penalizer.test.ts
antomor: >do you mean that we cannot test an edge case of, say, presenting a claim within 1 second of it being expired?

yep, I meant exactly that, and I have already tried to set values as close as possible, without impacting the developer experience (e.g. I don't want failing tests because of the time)

test/penalizer/Penalizer.test.ts
antomor: Agreed, thanks for the suggestions. It's not required when the claim call is rejected.

test/penalizer/Utils.ts
```diff
@@ -113,13 +113,13 @@ export class RelayHelper {
     )
   }
 
-  createReceipt (relayRequest: RelayRequest, signature: string): CommitmentReceipt {
+  createReceipt (relayRequest: RelayRequest, signature: string, time: number = Math.floor(Date.now() / 1000)): CommitmentReceipt {
```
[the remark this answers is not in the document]
antomor: > also, while we're at it, is the default value set to "now" so that claims are valid by default?
>
>although if this is the case, what about the slack time of 15 seconds?
>
> would appreciate your thoughts on this

Yep, ATM now is a valid value, but it actually depends on how this function is called, or better **when** it's called and on the use case. i.e. do we want to create a receipt that will be accepted? Do we want to create a receipt that will be rejected because of the time? that's why I added the possibility to specify the time 😅

antomor:
> i've left some comments down below, but nothing major.

Thanks for the comments/suggestions, very appreciated.

> now that we have more than 1 instance of a successful penalization, shouldn't we check to see what happens if we submit 2 valid claims but stake hasn't been added back between them?

Of course, we should test for it, I am going to add it in the next commit, good point!

test/penalizer/Utils.ts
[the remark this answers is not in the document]
mortelli: > > also, while we're at it, is the default value set to "now" so that claims are valid by default?
> > although if this is the case, what about the slack time of 15 seconds?
> > would appreciate your thoughts on this
> 
> Yep, ATM now is a valid value, but it actually depends on how this function is called, or better **when** it's called and on the use case. i.e. do we want to create a receipt that will be accepted? Do we want to create a receipt that will be rejected because of the time? that's why I added the possibility to specify the time sweat_smile

sounds good to me. it would be more confusing for this function to return a receipt which is rejected by default, instead of accepted

test/penalizer/Penalizer.test.ts
mortelli: then we agree. resolving this

test/penalizer/Penalizer.test.ts
mortelli: thanks, i think the new name is better :+1:

test/penalizer/Penalizer.test.ts
mortelli: > we could set lower values and take the "risk" that sometimes the tests fail. WDYT?

no, let's leave it as it is. i just did not expect to need 60 seconds, but i guess all of the test infrastructure (plus machine details) could justify this

test/penalizer/Penalizer.test.ts
mortelli: > I think that the line is required in order to make tests independent from each other. I mean, I believe that the current test shouldn't depend on other tests, but all the required operations should be performed in the `before`/`beforeEach` statements. Ideally, I should be able to run each test alone, without the need of running other tests to satisfy pre-conditions. Maybe we should adjust the other tests to avoid that also. WDYT?

in principle, i agree entirely. in practice though, it's not what's happening. as it is, line `278` is currently overwriting the hub value with its current value (resulting in no change). this is because `penalizer` is a global variable—available to all tests—as are the other contracts (as well as a couple other variables defined from line `28` onwards). 

while tests should be ideally independent from each other, i'm not sure it's worth the effort of, for example, creating a new environment (line `43`), deploying all the contracts, funding, etc., for each test case that we want to run. although doing this would indeed increase independence, the cost would be execution time, and _possibly_ readability and maintainability.

in the end, we're trying to strike a balance between independence and execution time/code complexity, and i think this can be achieved in more than one way. however, if we want to set the hub from scratch for each test, then we should be also deploying the contracts each time, and this would mean creating a new environment each time too. i think this might be overkill.

test/penalizer/Penalizer.test.ts
mortelli: @antomor 

i've created a [PR](https://github.com/antomor/rif-relay/pull/1) with this change; please check it out (and change and merge it as you see fit)

mortelli (approved):
i am approving although we have a couple of minor threads open.

thanks @antomor for your changes, and please make sure the tests run OK before merging

test/penalizer/Penalizer.test.ts
antomor: Yep, I agree that it could be overkill however, I tend to split the initialization code and the business logic of what I want to test. I think it could be acceptable to avoid an expensive initialization for the sake of optimization (time/computational cost), but I would like to be sure that changing a single test doesn't impact the execution of the other tests.
In my opinion, I would prefer to call that method on the initialization and "unset" it just before that single unit test (if possible). I know it's a subjective point of view, so no problem for me if you prefer to remove it from here, I get your point.

test/penalizer/Penalizer.test.ts
antomor: > @antomor
> 
> i've created a [PR](https://github.com/antomor/rif-relay/pull/1) with this change; please check it out (and change and merge it as you see fit)

Thanks a lot, that saved me time 🙏🏽

antomor:
> i am approving although we have a couple of minor threads open.
> 
> thanks @antomor for your changes, and please make sure the tests run OK before merging

@mortelli Thank you for the suggestions (and PR) 😊

test/penalizer/Penalizer.test.ts
mortelli: we had a call with @antomor and agreed to fix this as shown in [this commit](https://github.com/rsksmart/rif-relay/pull/165/commits/18124f3c8d3cab3b79adab31b648e7ca0500f200)

PR #167, opened 2021-10-07
PP-80: Add some tests to verify that a tx cannot be submitted twice

antomor:
This PR is the result of the analysis done upon a PR comment: https://github.com/rsksmart/rif-relay/pull/144/files#r699562838.
We want to be sure that a transaction already submitted cannot be re-submitted.
As described in the PR comment, the RelayServer already has some checks to avoid that, but there were no unit tests to cover these cases.
This PR includes the following unit tests:
- the same tx submitted twice
- a submitted transaction is tampered with the goal of being submitted.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,7 +5,7 @@ parameters:
         default: "rsksmart/rskj.git"
     version:
         type: string
-        default: "3.1.0-SNAPSHOT"
+        default: '3.2.0-SNAPSHOT'
     branch:
         type: string
         default: "master"
@@ -100,11 +100,18 @@ jobs: # a collection of steps
           name: Tests without DB reset Set B
           command: |
                     java -Dminer.client.autoMine=true -Drsk.conf.file=~/gls/rsknode/node.conf -Dminer.minGasPrice=1 << pipeline.parameters.hardfork >> -cp ~/rsksmart/rskj/rskj-core/build/libs/rskj-core-<< pipeline.parameters.version >>-all.jar co.rsk.Start --regtest --reset nohup &
-                    until nc -z 127.0.0.1 4444
-                    do
-                      echo "Waiting for RskJ..."
+                    _i=0
+                    while [ $_i -lt 30 ]; do
+                      curl -Ss -H "Content-type: application/json" \
+                          -d '{"jsonrpc":"2.0","method":"eth_blockNumber","id":"boot-test"}' \
+                          http://127.0.0.1:4444/ 2>/dev/null | grep -q result && break
                       sleep 1
+                      _i=$((_i + 1))
                     done
+                    if [ $_i -eq 30 ]; then
+                      echo "No response from rskj after 30 seconds; aborting..." >&2
+                      exit 1
+                    fi
                     npx truffle test --network rsk test/Verifiers.test.ts
                     npx truffle test --network rsk test/RelayHubPenalizations.test.ts
                     npx truffle test --network rsk test/RelayHubRegistrationsManagement.test.ts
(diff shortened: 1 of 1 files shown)


test/relayserver/RelayServer.test.ts
```diff
@@ -365,18 +366,55 @@ contract('RelayServer', function (accounts) {
...
+    it('should fail if the same relay transaction is submitted twice', async () => {
+      const relayTxRequest = await createAndVerifyRelayTransaction()
+      await expectRelayTransactionToFail(relayTxRequest, 'Unacceptable relayMaxNonce')
+    })
+
+    describe('with the goal of submitting a relay transaction twice', () => {
+      it('should fail if the relayMaxNonce is manually tampered ', async () => {
+        const relayTxRequest = await createAndVerifyRelayTransaction()
+        relayTxRequest.metadata.relayMaxNonce = relayTxRequest.metadata.relayMaxNonce + 1
+        await expectRelayTransactionToFail(relayTxRequest, 'revert nonce mismatch')
```
mortelli, line 408: i am wondering why this doesn't fail with a `revert signature mismatch`, since all of these steps are included in the next test.

is it because it's missing the `relayTxRequest.relayRequest.request.nonce` assignment?

so in a way this would be tampering with the nonce erroneously, right? the next test does it correctly, only the signature doesn't match.

if this is the case, i am ok with it

mortelli (approved):
LGTM, some minor comments.

non-blocker alternative for signatures:

both `createAndVerifyRelayTransaction` and `expectRelayTransactionToFail` use `env.relayServer.createRelayTransaction(request)` but their signatures differ quite a bit, which was a little bit confusing to me. 

also, the former returns a relay request instead of a transaction, right? this could arguably make its name misleading.

so again, not a blocker, but you could also do:
- `expectRelayTransactionToSucceed(request: RelayTransactionRequest): Promise<void>`: this contains lines `371` to `376`
- `expectRelayTransactionToFail(request: RelayTransactionRequest, msg: string): Promise<void>` (unchanged)

this would require you to build the correct requests for the first case (which is only 1 line: `370`) only this time you wouldn't be returning them.

i think this would make these functions a bit more symmetric

test/relayserver/RelayServer.test.ts
```diff
@@ -365,18 +366,55 @@ contract('RelayServer', function (accounts) {
...
       // TODO: add asserts here!!!
+      return req
+    }
+
+    async function expectRelayTransactionToFail (request: RelayTransactionRequest, msg: string): Promise<void> {
+      try {
+        await env.relayServer.createRelayTransaction(request)
+        fail("relayServer.createRelayTransaction() should have failed, but it didn't")
+      } catch (error) {
+        const errorMsg = (error as Error).message
```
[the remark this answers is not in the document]
antomor: No, it actually doesn't.  In typescript type castings are remove upon compilation, so it doesn't raise an error.
[Here](https://www.typescriptlang.org/play?ts=3.8.3#code/MYewdgzgLgBApgJwTAvDMcDuMCiSQIAUA5ALYCeeCBxAlANwBQjoksiCATKjJ81AnIwA3oxjiYUABbVsHRgF8YwAIZRgUwhwK0RYiawggANnAB0xkAHMt+ZCoi47AGhjEo5AA5xlDqAEswKxhSOAgIFSs4On1xQ1hSCGC0W2p7RyodM1DwyLhY5XAjUwtrQkSrWkV+QT0JSRkQOSQ+JVV1TW0EXVF6wxNzSxsumAcnNNd3Lx9VaEDgnIiomL6ihKSeVIJRjLtabLCl-NXIAdKbCqqFIA) you can find a typescript playground.

test/relayserver/RelayServer.test.ts
antomor: >is it because it's missing the `relayTxRequest.relayRequest.request.nonce` assignment?
yes, indeed:
1. the `relayMaxNonce` is checked by the `RelayServer` -> "Unacceptable relayMaxNonce"
2. if we try to temper it, then the transaction fails because the `SmartWallet` check for the `nonce` -> "nonce mismatch"
3. while if we temper the nonce also, the transaction fails because the `SmartWallet` check for the signature -> "signature mismatch"

PR #168, opened 2021-10-07
Removing ^Compatible with version from package.json

wilsoniovlabs:


antomor (changes requested):
I think we applied two different strategies to solve one single problem.
After reading some controversial opinions on pushing/not pushing the lock files (`package-lock.json` or `yarn.lock`), I think that the best option would be:
- if the repo is not supposed to be used as a dependency of another repo, hence it's not a library, we could push the lock files safely because it doesn't have any side-effect (I would suggest pushing the `package-lock.json` file and adding `yarn.lock` in the `.gitignore` to avoid confusion)
- if the repo is a library, I would suggest fixing the dependencies by using the `package.json` file,  in particular, removing the `^` from the dependencies that don't respect the [semver system](https://semver.org/). If a package **does respect** semantic versioning, it wouldn't be necessary to set a specific version.

Further details can be found on the related Jira task.

package.json
```diff
@@ -54,7 +54,7 @@
     "@types/lodash": "^4.14.161",
     "@types/minimist": "^1.2.0",
     "@types/nedb": "^1.8.11",
-    "@types/node": "^13.0.0",
+    "@types/node": "^13.13.4",
```
mortelli, line 57: could you please explain the reason behind this change?

wilsoniovlabs: The version 13.0.0 not exists in the NPM package versions https://www.npmjs.com/package/@types/node.

antomor:
LGTM

The problem I encountered was due to my own environment. I was using npm v6, and it tries to generate a v1 `package-lock.json` while the one in the repo is a v2 `package-lock.json`. After switching to npm v7, I installed it successfully.

PR #172, opened 2021-10-14
PP-42: Reduce gas consumption on RelayHub and Penalizer

antomor:
- Replace `encodeWithSignature` with `encodeWithSelector`
- pre-calculate function selectors
- restore `(signature)` statement

### DIFF
--- a/contracts/RelayHub.sol
+++ b/contracts/RelayHub.sol
@@ -200,6 +200,9 @@ contract RelayHub is IRelayHub {
         EnvelopingTypes.RelayRequest calldata relayRequest,
         bytes calldata signature
     ) external override returns (bool destinationCallSuccess){
+        /* statement originally present in the GSN repo; 
+         * we left it here because of a small gas improvement noticed on tests execution.
+         */
         (signature);
         require(msg.sender == tx.origin, "RelayWorker cannot be a contract");
         require(


contracts/Penalizer.sol
```diff
@@ -15,6 +15,8 @@ import "./interfaces/IPenalizer.sol";
 contract Penalizer is IPenalizer, Ownable {
 
     string public override versionPenalizer = "2.0.1+enveloping.penalizer.ipenalizer";
+    // bytes4(keccak256("penalize(address,address)"))
+    bytes4 private constant PENALIZE_SELECTOR = 0xebcd31ac;
```
mortelli, line 19: _love_ this from a programmatic point of view. 

however, would an in-line value reduce costs even further?

contracts/RelayHub.sol
```diff
@@ -194,6 +196,7 @@ contract RelayHub is IRelayHub {
         EnvelopingTypes.RelayRequest calldata relayRequest,
         bytes calldata signature
     ) external override returns (bool destinationCallSuccess){
+        (signature);
```
mortelli, line 199: i know this was commented on the original PR, but i am baffled by it. do we have any idea why this statement reduces costs?

i don't even know how a programmer would even arrive at this :exploding_head:

mortelli:
@antomor i've left 2 additional comments on [the JIRA usse](https://rsklabs.atlassian.net/browse/PP-42)

contracts/Penalizer.sol
antomor: I have just tried your suggestion and I hadn't observed any gas improvement. For the sake of readability, I would define it here, WDYT?

mortelli: yes, better to leave it as it is :+1:

contracts/RelayHub.sol
mortelli: let's leave the spooky statement in, but maybe add an in-line comment to explain

mortelli:
```suggestion
        /* statement originally present in the GSN repo; 
         * we left it here because of a small gas improvement noticed on tests execution.
         */
        (signature);
```

PR #178, opened 2021-11-15
Fix: Signature verify when create wallet (not using RELAY)

patogallaiovlabs:
## Tests (BIG Disclaimer!!!)
- This has only been tested manually, no formal test developed nor modified in this PR so far, this is a draft.

## Context
When using the SmartWalletFactory contract without using the relay, directly from an EOA, the function "createUserSmartWallet" is called. This function checks if the signature belongs to the owner of the wallet and that the message signed is: `hash(<owner> + <recoverer> + <index>)`.

## Issue
Using Metamask/Nifty wallets at the moment of signing the message, a prefix is added so the final message would be:
`"\x19Ethereum Signed Message:\n" + msg.length + hash(<owner> + <recoverer> + <index>)`.
This is due to `eth_sign` is deprecated by those wallets:
- https://github.com/poanetwork/nifty-wallet/blob/master/CHANGELOG.md#399-2017-8-18
- https://github.com/MetaMask/metamask-extension/issues/2215

## Solution
So, in order to be compatible with this behaviour is that in this PR the verification of the signature (only in the case of direct deploy, without use of the RELAY mechanism) has been changed to the following:
- Message to sign in the **client** side is: `hash(<verifier_address> + <owner> + <recoverer> + <index>)`
- **Client** should use `presonal_sign` function (which prepends `"\x19Ethereum Signed Message:\n" + msg.length`).
- The recovery process in the Contract is: 
   - msg = `"\x19Ethereum Signed Message:\n" + msg.length + hash(<verifier_address>+<owner> + <recoverer> + <index>)`
   - `hash(msg).recover(signature)`

### DIFF
--- a/contracts/factory/CustomSmartWalletFactory.sol
+++ b/contracts/factory/CustomSmartWalletFactory.sol
@@ -112,6 +112,7 @@ contract CustomSmartWalletFactory is ICustomSmartWalletFactory {
         bytes calldata sig
     ) external override {
         bytes32 _hash = keccak256(abi.encodePacked(
+            address(this),
             owner,
             recoverer,
             logic,
--- a/contracts/factory/CustomSmartWalletFactory.sol
+++ b/contracts/factory/CustomSmartWalletFactory.sol
@@ -111,21 +111,17 @@ contract CustomSmartWalletFactory is ICustomSmartWalletFactory {
         bytes calldata initParams,
         bytes calldata sig
     ) external override {
-        bytes memory packed =
-            abi.encodePacked(
-                "\x19\x10",
-                owner,
-                recoverer,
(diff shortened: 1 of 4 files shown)


raullaprida:
@mortelli You can add me as reviewer once it is ready-for-review!

mortelli:
relevant PRs:
1. [contracts](https://github.com/anarancio/rif-relay-contracts/pull/17)
2. [tests](https://github.com/rsksmart/rif-relay/pull/183)

antomor:
These changes have been included in the [relevant PRs](https://github.com/rsksmart/rif-relay/pull/178#issuecomment-1012509212).

PR #192, opened 2022-02-18
PP-103: fix missing folder during test execution on MacOS

antomor:
## What
- Update package-lock.json
- Fix the execution of test against docker since that it was using the first file specified as test name, so that file was skipped from the test execution.
-  Wrap the deletion of the temporary folder `'/tmp/enveloping/test/server'` in a try/catch to avoid errors on MacOs


mortelli (approved):
thanks for updating the package-lock file :raised_hands:

PR #193, opened 2022-02-18
PP-103: fix missing folder on tests [develop]

antomor:
## What
- Fix the execution of the tests against docker since that it was using the first file specified as test name, so that file was skipped from the test execution.
-  Wrap the deletion of the temporary folder `'/tmp/enveloping/test/server'` in a try/catch to avoid errors on MacOs

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -31,12 +31,12 @@ commands:
                       ./gradlew clean build -x test
 
             - restore_cache: # special step to restore the dependency cache
-                  key: dependency-cachev6-{{ checksum "package.json" }}-npm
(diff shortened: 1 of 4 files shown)


mortelli (commented):
thank you very much for this correction :+1: 

i am not approving due to the fact that CI is failing (now's the time that i'm reminded that it would be cool to have CI on the other repos too)

the cache change does not seem to work so let's have a look on monday.

thanks

mortelli (approved):

antomor:
> (now's the time that i'm reminded that it would be cool to have CI on the other repos too)
Yes, definitely, but before doing that, we need to change repo ownership

PR #196, opened 2022-04-19
Add some tests to estimate the gas overhead to execute transfer txs

antomor:
## What
- Add some tests that log the gas overhead when executing a transfer transaction
- We skipped them for the moment; the repo already includes some tests for the same purpose and all of them are skipped (they don't test the functionalities, but they are needed for analysis purposes)

## Why
- We want to have a clear idea about the gas overhead introduced by RIF relay when executing different transactions.

### DIFF
--- a/test/RelayHub.test.ts
+++ b/test/RelayHub.test.ts
@@ -982,7 +982,7 @@ contract(
                         );
 
                         console.log(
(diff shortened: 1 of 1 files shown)


antomor:
Tests included and refactored with #217

PR #197, opened 2022-04-19
[CI][Fix] Update RSKj version

antomor:
## What
- update the RSKj version used to run the tests

## Why
- we need to change the RSKj version used each time a new version is released. We have already planned the change of this strategy to avoid this.


antomor:
The tag 4.1.0 was introduced by mistake, so this's not valid anymore.

Issue #179, opened 2021-12-01
Remove transfer of balance from SmartWallet to ownerEOA


raullaprida:
As it has been discussed during UX meetings. It would be more consistent (simpler, in UX terms) to leave the native cryptocurrency balance in the SmartWallet instead of transferring it to the ownerEOA after each execution.

That involves removing the following piece of code:

` //If any balance has been added then trasfer it to the owner EOA`
 `          uint256 balanceToTransfer = address(this).balance;`
 `           if ( balanceToTransfer > 0) {`
 `                 //can't fail: req.from signed (off-chain) the request, so it must be an EOA...`
 `                 payable(req.from).transfer(balanceToTransfer);`
 `            }`

And also modify the `directExecute` function so that it uses the SmartWallet's balance  (if value needs to be sent to the destination contract) instead of expecting a sufficient `msg.value` in the transaction 

For example:

`  function directExecute(address to, uint256 value, bytes calldata data) external override payable returns (`
            `bool success, bytes memory ret)`
    `{`
       ` _verifyOwner();`
       ` (success, ret) = to.call{value: value}(data);`
   ` }`

It is the same logic, but the actual numeric value to send is received as a parameter, and it is expected to be available in the SmartWallet's balance. (The current implementation uses `msg.value`)

The `directExecute` can remain `payable` anyway.

mortelli:
[PP-136](https://rsklabs.atlassian.net/browse/PP-136)

PR #198, opened 2022-05-24
PP-136/smart wallet transfer

franciscotobar:
## What

- Testing was updated to match the changes on the rif-relay-contracts

## Refs

- [Issue](https://github.com/rsksmart/rif-relay/issues/179)
- [rif-relay-contracs] https://github.com/rsksmart/rif-relay-contracts/pull/28


antomor (approved):
LVGTM!

antomor (changes requested):
After checking the dapp, I realized that it may be useful to add a test that checks that the `directExecute` fails if the destination function isn't payable. WDYT?

PR #199, opened 2022-05-26
PP-136/smart wallet transfer

franciscotobar:
## What

Removed the process where native cryptocurrency balance is transfer back to the owner EOA

## Why

It would be more consistent (simpler, in UX terms) to leave the native cryptocurrency balance in the SmartWallet instead of transferring it to the owner EOA after each execution.

## Refs

- [Jira issue](https://github.com/rsksmart/rif-relay/issues/179)

### DIFF
--- a/contracts/test/ProxyCustomLogic.sol
+++ b/contracts/test/ProxyCustomLogic.sol
@@ -34,7 +34,7 @@ contract ProxyCustomLogic is IWalletCustomLogic {
         require(success, "call failed");
     }
 
-    function directExecute(address to, bytes calldata data) override external payable returns (
+    function directExecute(address to, uint256 value, bytes calldata data) override external payable returns (
         bytes memory ret  
     ) {  
         emit LogicCalled();  
--- a/contracts/test/ProxyCustomLogic.sol
+++ b/contracts/test/ProxyCustomLogic.sol
@@ -39,7 +39,7 @@ contract ProxyCustomLogic is IWalletCustomLogic {
     ) {  
         emit LogicCalled();  
         bool success;              
-        (success, ret) = to.call{value: msg.value}(data);
+        (success, ret) = to.call{value: value}(data);
         require(success, "call failed");
 
     }
--- a/rsknode/Dockerfile
+++ b/rsknode/Dockerfile
@@ -23,8 +23,8 @@ RUN ./gradlew --no-daemon clean build -x test
 
 FROM openjdk:8-jdk-slim-buster
 
-ARG rskj_codename=SNAPSHOT
-ARG rskj_version=3.2.0
+ARG rskj_codename=IRIS
+ARG rskj_version=3.3.0
 
 WORKDIR /usr/local/rskj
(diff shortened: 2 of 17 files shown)


contracts/test/ProxyCustomLogic.sol
```diff
@@ -34,7 +34,7 @@ contract ProxyCustomLogic is IWalletCustomLogic {
         require(success, "call failed");
     }
 
-    function directExecute(address to, bytes calldata data) override external payable returns (
+    function directExecute(address to, uint256 value, bytes calldata data) override external payable returns (
         bytes memory ret  
     ) {  
         emit LogicCalled();
```
antomor, line 40: I think we should change line#42 to replace `msg.value` with `value` so it would be something like:
```
(success, ret) = to.call{value: value}(data);
```

rsknode/Dockerfile
```diff
@@ -8,9 +8,9 @@ RUN apt-get update -y && \
...
 RUN gpg --finger 1A92D8942171AFA951A857365DECF4415E3B8FA4
 
-ARG rskj_codename=SNAPSHOT
-ARG rskj_version=3.2.0
-ARG working_branch=master
+ARG rskj_codename=IRIS
+ARG rskj_version=3.3.0
+ARG working_branch=IRIS-3.3.0
 
 RUN git clone --single-branch --depth 1 --branch "${working_branch}" https://github.com/rsksmart/rskj.git /code/rskj
```
antomor, line 16: I think we should change also lines 26 and 27 to:
```
ARG rskj_codename=IRIS
ARG rskj_version=3.3.0
```

antomor (changes requested):
@franciscotobar thank you for these changes. Apart from the comments I left, I think we should apply the formatting rules to avoid the CI/CD pipeline fails. Good job 💪🏼

rsknode/Dockerfile
antomor: @franciscotobar I think this is still valid, isn't it?

antomor (changes requested):
@franciscotobar Thank you for adding the tests. Just a minor issue is still open

test/smartwallet/SmartWallet.test.ts
```diff
@@ -550,26 +548,38 @@ options.forEach(element => {
...
+
           assert.equal(logs.length, 1, 'TestRecipient should emit')
           assert.equal(logs[0].args.origin, otherAccount, 'test "from" account is the tx.origin')
           assert.equal(logs[0].args.msgSender, sw.address, 'msg.sender must be the smart wallet address')
 
           assert.equal((await sw.nonce()).toString(), initialNonce.toString(), 'direct execute should NOT increment nonce')
         })
 
+        it('should fail if executed function is non-payable', async () => {
+          const func = recipient.contract.methods.emitMessage('hello').encodeABI()
```
antomor, line 564: Just checked and in my understanding, this function shouldn't fail because it's payable. What Am I missing?

antomor (approved):
LGTM!

PR #200, opened 2022-06-13
patch: installs rif libs from npm and removes dist folder

jurajpiar:
## What

- removes dist folder and old scripts
- adds build to prepare script
- changes rif deps from github to npm (except for the rif-relay-server)

## Why

- Reducing inconsistencies in the rif-relay projects

## Refs

- https://rsklabs.atlassian.net/browse/PP-166
- https://rsklabs.atlassian.net/browse/PP-74


antomor (changes requested):
@jurajpiar thank you for these changes.
I noticed that the pipeline is failing and it may be related to the update of the `package-lock.json` file (maybe because it's the lockfile version 2, but it's a guess). Could you verify, please?

PR #209, opened 2022-07-15
PP 271: Fix configurator tests

antomor:
## What

- Fix configurator tests

## Why
- The expected error message of one of the tests of the configurator is wrong


david-iov (approved):
LGTM

PR #211, opened 2022-07-21
chore: cleans up ignores + ups lock file

jurajpiar:
### what:
removes a number of unnecessary entries in gitignore
updates package-lock
stops copying contracts dir from dependency and instead adds it to truffle config

### why:
to clean up a little ;]


antomor:
We're going to update all the repo dependencies now

PR #214, opened 2022-08-01
feature/PP-204

franciscotobar:
## What

- Validation for smart wallet address to verify if it belongs to the EOA and its deployed

## Why

- To be able to import a smart wallet

## Refs

- [Jira issue](https://rsklabs.atlassian.net/browse/PP-204)
- https://github.com/rsksmart/rif-relay-common/pull/25
- https://github.com/rsksmart/rif-relay-client/pull/23


ironFe93 (commented):
Do you think we could use assert.rejects() instead of try-catch? https://stackoverflow.com/questions/54247064/mocha-node-how-to-use-assert-rejects

AndresQuijano (approved):
LGTM

PR #219, opened 2022-11-16
PP-574: analyse gas consumption

antomor:
## What

- Refactor the gas consumption tests to measure transactions with and without token payment without requiring manual changes.
- Highlight the gas overhead calculated during the test execution.
image

## Why

- In order to execute those tests before, we had to comment-out some parts of the code.


franciscotobar (commented):
LGTM

franciscotobar (approved):

PR #233, opened 2023-01-17
PP-642/migrate-test-utils-server

franciscotobar:
## What

- Migrate utility files used by the server

## Why

- Necessary to integration test

## Refs

- [PP-642](https://rsklabs.atlassian.net/browse/PP-642)
- https://github.com/rsksmart/rif-relay-client/pull/69
- https://github.com/rsksmart/rif-relay-server/pull/75


test/relayserver/ServerTestEnvironments.ts
```diff
@@ -0,0 +1,233 @@
...
+  );
+};
+
+const relayTransaction = async (
+  envelopingRequest: UserDefinedEnvelopingRequest,
+  relayServer: RelayServer,
+  relayClient: RelayClient,
+  assertRelayed = true
+) => {
+  const hubInfo = relayServer.pingHandler();
```
franciscotobar, line 171: should we change the name of ´pingHandler´ to follow the RelayClient one?

test/relayserver/ServerTestEnvironments.ts
```diff
@@ -0,0 +1,233 @@
...
+      smartWalletFactoryAddress: factory.address,
+    },
+    app: {
+      devMode: true,
+      checkInterval: 10,
+      logLevel: 5,
+    },
+  });
+
+  const relayServer = new RelayServer(dependencies);
```
franciscotobar, line 126: We cannot set the provider from the server, it get the provider from ´getDefaultProvider´, but we need to use the hardhat one.

test/TestUtils.ts
```diff
@@ -0,0 +1,394 @@
...
+
+const revertSnapshot = async (snapshotId: string) => {
+  return (await provider.send('evm_revert', [snapshotId])) as boolean;
+};
+
+const verifyRelayServerStatus = async (
+  url: string,
+  attempts: number,
+  interval = 1000,
+  isReady = false
```
antomor, line 139: Why do we need the `isReady` param?

test/TestUtils.ts
```diff
@@ -0,0 +1,394 @@
...
+
+      return response;
+    } catch (e) {
+      /* empty */
+    }
+  }
+
+  expect(response, "can't ping server").to.be.ok;
+
+  throw Error("can't ping server");
```
antomor, line 162: I think we could either expect something or throw the error. I think that using both may be redundant

test/relayserver/ServerTestEnvironments.ts
antomor: We may need to use a similar approach to what we used in the client.

test/relayserver/ServerTestEnvironments.ts
```diff
@@ -0,0 +1,233 @@
...
+  if (workdir != null) {
+    return new KeyManager(1, workdir);
+  } else {
+    return new KeyManager(1, undefined, utils.randomBytes(32).toString());
+  }
+};
+
+const createRelayHttpRequest = async (
+  envelopingRequest: UserDefinedEnvelopingRequest,
+  relayClient: RelayClient,
```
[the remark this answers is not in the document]
antomor: This was a request from one of our internal users, but the problem is that in that way the server needs to use that verifiers and smartwallet only, and that could be a limitation in a multi-tenant environment. Let's pass the relayClient for now.

test/relayserver/ServerTestEnvironments.ts
antomor: Yes, we should do it for consistency

antomor (changes requested):

test/relayserver/ServerTestEnvironments.ts
franciscotobar: I think that we can stub the getProvider from the server, to change that, but I need to test it.

test/TestUtils.ts
franciscotobar: because, we need to know first if the server is receiving requests and after staking check if its ready for transactions. Therefor I need to ping the server without being ready to get the managerAddress and then I need to ping it again once its ready after doing the stake process.

test/TestUtils.ts
franciscotobar: I would prefer to throw error, I think that I have this approach on multiple places, Ill update it.

test/relayserver/ServerTestEnvironments.ts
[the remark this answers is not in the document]
franciscotobar: no sorry, no need for discussion, I was mixing things and got confused but I have it clear now.

package.json
```diff
@@ -79,14 +79,18 @@
     "package.json": "npx sort-package-json"
   },
   "dependencies": {
-    "@rsksmart/rif-relay-contracts": "github:rsksmart/rif-relay-contracts#feature/PP-278/PP-329/upgrade_contracts"
+    "@rsksmart/rif-relay-client": "github:rsksmart/rif-relay-client#PP-642/test-utils-server",
+    "@rsksmart/rif-relay-contracts": "github:rsksmart/rif-relay-contracts#feature/PP-278/PP-329/upgrade_contracts",
+    "@rsksmart/rif-relay-server": "github:rsksmart/rif-relay-server#PP-642/test-utils-server",
+    "config": "^3.3.8"
```
franciscotobar, line 85: should I move this to devDependencies

package.json
```diff
@@ -95,10 +99,11 @@
     "eslint-config-prettier": "^8.6.0",
     "eslint-plugin-mocha": "^10.1.0",
     "eslint-plugin-prettier": "^4.2.1",
-    "ethers": "^5.7.2",
+    "ethers": "5.7.0",
     "hardhat": "^2.12.5",
```
jurajpiar, line 103: should be in dependencies otherwise the tests wont work in CI, right?

jurajpiar, line 103: should be in dependencies?

package.json
```diff
@@ -95,10 +99,11 @@
...
     "husky": "^8.0.3",
     "lint-staged": "^13.1.0",
-    "prettier": "^2.8.1"
+    "prettier": "^2.8.1",
+    "ts-node": "^10.9.1"
```
jurajpiar, line 107: should be in dependencies?

config/default.json
```diff
@@ -0,0 +1,19 @@
...
+        "devMode": true,
+        "customReplenish": false,
+        "logLevel": 1,
+        "workdir": "/Users/borja/iovlabs/projects/temp/PP-358/environment"
+    },
+    "blockchain": {
+        "gasPriceFactor": 1,
+        "rskNodeUrl": "http://127.0.0.1:4444"
+    }
+}
```
jurajpiar, line 19: put this in `./config/local.json(5)` please

test/relayserver/ServerTestUtils.ts
```diff
@@ -0,0 +1,86 @@
+import { TransactionReceipt } from '@ethersproject/providers';
+import {
+  RelayHubInterface,
+  RelayHub__factory,
+} from '@rsksmart/rif-relay-contracts';
+import { ManagerEvent } from '@rsksmart/rif-relay-server';
+import { BigNumberish, constants } from 'ethers';
+import { expect } from 'chai';
+import { ethers as hardhat } from 'hardhat';
```
jurajpiar, line 9: why do we need to rename it? It's the same object in runtime as directly from ethers library with added helper functions.

franciscotobar: I just followed the same structure that was usted in the relay-contracts, buy I could keep the same name.

config/default.json
franciscotobar: I'm still waiting for your PR to me merged to do that :)

test/TestUtils.ts
```diff
@@ -0,0 +1,396 @@
...
+    maxWorkerCount,
+    minimumUnstakeDelay,
+    minimumStake,
+    minimumEntryDepositValue
+  );
+};
+
+const createSmartWalletFactory = async (
+  template: IForwarder,
+  isCustom = false
```
ironFe93, line 221: I agree with this isCustom param :)

package.json
franciscotobar: That is what confuse me. I understand that config needs to be available in a production, but this repo is for testing purposes, is there ever going to be used in production?

test/relayserver/ServerTestEnvironments.ts
franciscotobar: I was able to stub it.

```typescript
Sinon.stub(ethers, 'getDefaultProvider').returns(hardhat.provider);
```

Should I stub it, or create a function to setup it?

antomor: I think we could stub it for now

test/relayserver/ServerTestEnvironments.ts
antomor: Should we include this change in the [server PR](https://github.com/rsksmart/rif-relay-server/pull/75)

test/TestUtils.ts
antomor: Now I see what you mean, but what about creating an additional function for that specific case?
```ts
const getServerStatus = (url: string, client = new HttpClient(new HttpWrapper(undefined, 'silent'))) => client.getChainInfo(url);

// TODO: this function could be renamed to something like waitForServerToBeReady
const verifyRelayServerStatus = async (
  url: string,
  attempts: number,
  interval = 1000
): Promise<HubInfo> => {
  let response: HubInfo | undefined;
  const httpClient = new HttpClient(new HttpWrapper(undefined, 'silent'));

  for (let i = 0; i < attempts; i++) {
    try {
      console.log('sleep before cont.');
      await sleep(interval);
      response = await getServerStatus(url, httpClient);

      if (!response.ready) {
        continue;
      }

      return response;
    } catch (e) {
      /* empty */
     // I would log something here (I'm not a fan of empty catch blocks)
    }
  }

  throw Error("can't ping server");
};
```
So, instead of using it with the false flag, you could simply call `getServerStatus`. WDYT

test/relayserver/ServerTestUtils.ts
antomor: > not a big deal, but why do we need to rename it? It's the same object in runtime as directly from ethers library with added helper functions.

I agree that unless strictly necessary, I'd avoid renaming, because I find it a bit confusing.

test/relayserver/ServerTestEnvironments.ts
```diff
@@ -0,0 +1,227 @@
...
+    workdir: serverWorkdirs?.workdir ?? workdir,
+  });
+
+  const dependencies: ServerDependencies = {
+    txStoreManager,
+    managerKeyManager,
+    workersKeyManager,
+  };
+
+  const relayHub = await deployRelayHub(penalizerAddress, {});
```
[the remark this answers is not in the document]
antomor: Sorry, but apart from the configuration, I don't see any difference between the hub and the verifiers. Could you please elaborate a bit more?

test/relayserver/ServerTestEnvironments.ts
franciscotobar: yes Ill include it, sorry I was waiting for Jurajs PR, but Ill do it today, and then rebase Jurajs PR.

test/relayserver/ServerTestEnvironments.ts
franciscotobar: Ill stub it, during the process of the getInstance from the server.

test/TestUtils.ts
franciscotobar: I thought about it, but was trying to avoid create duplication, but we could do that. Ill update it and push it.

test/relayserver/ServerTestUtils.ts
franciscotobar: Ill update it

test/relayserver/ServerTestEnvironments.ts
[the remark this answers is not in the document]
franciscotobar: The verifiers are deployed outside this function and the address are passed, therefore the responsibility of creating verifiers its from the one that is going to use this function. Right now, the responsibility of deploying the relayHub its from this function. Take in consideration that we can get the relayHubAddress at any time from the server but we cannot get the verifiers, therefore I decided to not deploy it here, since we will not have access to it.

test/relayserver/ServerTestEnvironments.ts
franciscotobar: updated

test/relayserver/ServerTestEnvironments.ts
franciscotobar: updated

test/TestUtils.ts
franciscotobar: updated

test/TestUtils.ts
```diff
@@ -0,0 +1,426 @@
...
+
+  await relayHub.stakeForAddress(relayManagerAddress, delay || 2000, {
+    from: relayOwner,
+    value: stake || utils.parseEther('1'),
+  });
+
+  const { relayWorkerAddress } = await waitForFunction(
+    getServerReady,
+    url,
+    'timed out waiting for relay to get staked and registered',
```
jurajpiar, line 107: misleading in relationship to line 86: "can't ping server". suggests different approaches, though it's the same server, endpoint and timeout.

test/TestUtils.ts
```diff
@@ -0,0 +1,426 @@
...
+  );
+
+  const proc: ChildProcessWithoutNullStreams & { alreadyStarted?: number } =
+    childProcess.spawn('node', [runServerPath]);
+
+  const { relayManagerAddress } = await waitForFunction(
+    getServerStatus,
+    url,
+    "can't ping server",
+    3
```
jurajpiar, line 87: why 3?

test/TestUtils.ts
```diff
@@ -0,0 +1,426 @@
...
+    from: relayOwner,
+    value: stake || utils.parseEther('1'),
+  });
+
+  const { relayWorkerAddress } = await waitForFunction(
+    getServerReady,
+    url,
+    'timed out waiting for relay to get staked and registered',
+    25,
+    500
```
jurajpiar, line 109: why 25 and 500?

test/TestUtils.ts
```diff
@@ -0,0 +1,426 @@
...
+      }
+
+      return response;
+    } catch (e) {
+      console.log('getChainInfo failed');
+    }
+  }
+
+  throw Error(errorMessage);
+};
```
jurajpiar, line 166: I'd rename and simplify to:

```suggestion
const waitForHubInfoProperty = async (
  url: string,
  hubInfoProp: keyof HubInfo,
  attempts: number,
  interval = 1000,
  customErrorMessage?: string,
): Promise<HubInfo> => {
  const client = new HttpClient(new HttpWrapper(undefined, 'silent'));
  for (let i = 0; i < attempts; i++) {
    try {
      console.info('sleep before cont.');
      await sleep(interval);

      const hubInfo: HubInfo | undefined = await client.getChainInfo(url);

      if (hubInfo?.[hubInfoProp]) {
        continue;
      }

      return hubInfo;
    } catch (error: unknown) {
      console.error(`Failed to call getChainInfo on ${url}:`, error);
    }
  }

  throw Error(`${customErrorMessage && customErrorMessage + ': '}HubInfo poll expired and ${hubInfoProp} property was not aquired from ${url}.getChainInfo`);
};
```

as it seems to better capture the nature of what it actually does.

test/TestUtils.ts
```diff
@@ -0,0 +1,426 @@
...
+const increaseBlockchainTime = async (time: number): Promise<void> => {
+  await provider.send('evm_increaseTime', [time]);
+  await evmMine();
+};
+
+const createSnapshot = async (): Promise<string> => {
+  return (await provider.send('evm_snapshot', [])) as string;
+};
+
+const revertSnapshot = async (snapshotId: string) => {
```
jurajpiar, line 191:
```suggestion
const revertEvmSnapshot = async (snapshotId: string) => {
```

test/TestUtils.ts
```diff
@@ -0,0 +1,426 @@
...
+  return (await provider.send('evm_revert', [snapshotId])) as boolean;
+};
+
+// An existing account in RSKJ that have been depleted
+const getExistingGaslessAccount = async (): Promise<Wallet> => {
+  const gaslessAccount = new Wallet(
+    '0x082f57b8084286a079aeb9f2d0e17e565ced44a2cb9ce4844e6d4b9d89f3f595',
```
jurajpiar, line 198: how do we know that this address will always work?

jurajpiar (changes requested):
I didn't review everything as there is a number of comments already.

config/mainnet.json5
```diff
@@ -0,0 +1,5 @@
+{
+    blockchain: {
+        rskNodeUrl: 'https://public-node.rsk.co',
+    }
+}
```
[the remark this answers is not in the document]
franciscotobar: I dont think so, let me delete it :)

config/testnet.json5
```diff
@@ -0,0 +1,9 @@
+{
+    app: {
+      devMode: true,
+      logLevel: 3,
+    },
+    blockchain: {
+        rskNodeUrl: 'https://public-node.testnet.rsk.co',
+    }
+}
```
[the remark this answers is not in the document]
franciscotobar: I dont think so, let me delete it :)

config/default.json5
```diff
@@ -0,0 +1,85 @@
...
+    funds: '0.02', // amount of funds to set up
+    mnemonic: '', // mnemonic to use for unlocking the account parameter
+    privateKey: '', // private key to retrieve the account address from
+    signer: {}, // Signer object (only when used as a library)
+    hub: '',
+    gasPrice: 60000000,
+    relayUrl: '',
+    unstakeDelay: 1000,
+  },
+}
```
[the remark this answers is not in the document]
franciscotobar: we do, the `config`, package get the configuration from the root folder, so it will be using this files instead of the server one, unless we configure over here to use the ones that inside the server.

.gitignore
```diff
@@ -36,4 +36,8 @@ tsconfig.
 yarn.lock
 migrations
 contract-addresses.json
-typechain-types
\ No newline at end of file
+typechain-types
+txstore.db
+manager/keystore
+workers/keystore
+config/local.json5
```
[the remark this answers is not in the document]
franciscotobar: updated, sorry my bad

test/TestUtils.ts
```diff
@@ -0,0 +1,426 @@
...
+  RelayRequest,
+  relayRequestType,
+} from '@rsksmart/rif-relay-client';
+import { ethers } from 'hardhat';
+import { _TypedDataEncoder } from 'ethers/lib/utils';
+
+use(chaiAsPromised);
+
+const SERVER_WORK_DIR = '/tmp/enveloping/test/server';
+const ONE_FIELD_IN_BYTES = 32;
```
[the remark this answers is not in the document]
franciscotobar: updated

test/TestUtils.ts
franciscotobar: yes but the messages specifies what it happening, you are able to reach the server, but you the server was not ready for relaying transactions due to a long wait to be staked and registered

test/TestUtils.ts
franciscotobar: Ill put it as a constant, but it was the number previously set.

test/TestUtils.ts
franciscotobar: Ill put it as a constant, but it was the number previously set.

test/TestUtils.ts
```diff
@@ -0,0 +1,426 @@
...
+    await evmMine();
+  }
+};
+
+const increaseBlockchainTime = async (time: number): Promise<void> => {
+  await provider.send('evm_increaseTime', [time]);
+  await evmMine();
+};
+
+const createSnapshot = async (): Promise<string> => {
```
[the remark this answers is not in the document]
franciscotobar: updated

test/TestUtils.ts
franciscotobar: updated

test/TestUtils.ts
franciscotobar: actually Im assuming that is hardcoded in the node, I just followed what they had before, but I could change the approach to use the getAccount methods from the provider, but Im not pretty sure if its going to change something. @antomor what do you think?

test/relayserver/ServerTestUtils.ts
```diff
@@ -0,0 +1,86 @@
...
+
+export type ServerWorkdirs = {
+  workdir: string;
+  managerWorkdir: string;
+  workersWorkdir: string;
+};
+
+const provider = hardhat.provider;
```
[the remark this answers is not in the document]
franciscotobar: updated

test/TestUtils.ts
```diff
@@ -0,0 +1,426 @@
...
+  const response = await getServerStatus(url);
+
+  if (response.ready) {
+    return response;
+  }
+
+  return undefined;
+};
+
+const waitForFunction = async (
```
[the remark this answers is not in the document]
franciscotobar: updated

test/TestUtils.ts
```diff
@@ -0,0 +1,426 @@
...
+    contracts: {
+      relayHubAddress,
+    },
+    app: {
+      workdir: SERVER_WORK_DIR,
+      devMode: true,
+      checkInterval: 10,
+      logLevel: 5,
+    },
+  });
```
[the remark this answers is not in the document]
franciscotobar: node-config its not pointing to the server config folder, its pointing to the root config folder, can we discuss this after the daily?

package.json
franciscotobar: updated

package.json
franciscotobar: updated

config/mainnet.json5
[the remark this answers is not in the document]
franciscotobar: updated

config/testnet.json5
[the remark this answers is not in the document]
franciscotobar: updated

config/custom-environment-variables.json
```diff
@@ -0,0 +1,12 @@
...
+    "stake": "REGISTER_STAKE",
+    "funds": "REGISTER_FUNDS",
+    "mnemonic": "REGISTER_MNEMONIC",
+    "privateKey": "REGISTER_PRIVATE_KEY",
+    "hub": "REGISTER_HUB_ADDRESS",
+    "gasPrice": "REGISTER_GAS_PRICE",
+    "relayUrl": "REGISTER_RELAY_URL",
+    "unstakeDelay": "REGISTER_UNSTAKE_DELAY"
+  }
+}
```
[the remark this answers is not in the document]
franciscotobar: updated

.gitignore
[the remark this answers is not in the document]
franciscotobar: updated

test/TestUtils.ts
```diff
@@ -0,0 +1,426 @@
...
+  isDeployRequest,
+  RelayRequest,
+  relayRequestType,
+} from '@rsksmart/rif-relay-client';
+import { ethers } from 'hardhat';
+import { _TypedDataEncoder } from 'ethers/lib/utils';
+
+use(chaiAsPromised);
+
+const SERVER_WORK_DIR = '/tmp/enveloping/test/server';
```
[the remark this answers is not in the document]
franciscotobar: updated

test/TestUtils.ts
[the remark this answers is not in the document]
franciscotobar: updated

test/TestUtils.ts
franciscotobar: updated

test/TestUtils.ts
franciscotobar: we need to discuss it

config/regtest_ci.json5
```diff
@@ -0,0 +1,11 @@
+{
```
antomor, line 1: If we merge this, we probably don't need this file any more.

.gitignore
```diff
@@ -36,4 +36,8 @@ tsconfig.
 yarn.lock
 migrations
 contract-addresses.json
-typechain-types
\ No newline at end of file
+typechain-types
+txstore.db
+manager/keystore
+workers/keystore
```
antomor, line 42: I'm not sure we need these files here, because they're generated if and only if we set `workdir` with the current directory

hardhat.config.ts
```diff
@@ -1,19 +1,13 @@
...
-  sharedNetworkConfig.accounts = [PK];
-} else {
-  sharedNetworkConfig.accounts = {
-    mnemonic: MNEMONIC || DEFAULT_MNEMONIC,
-  };
-}
+const CONFIG_BLOCKCHAIN = 'blockchain';
+const CONFIG_RSK_URL = 'rskNodeUrl';
+
+const getRskNodeUrl = () =>
```
antomor, line 9: I like this 👍🏽

hardhat.config.ts
```diff
@@ -39,18 +33,13 @@ const config: HardhatUserConfig = {
   },
   networks: {
     regtest_ci: {
-      url: 'http://enveloping-rskj:4444',
+      url: getRskNodeUrl(),
```
antomor, line 36: If we merge #236, we don't need this any more.

test/TestUtils.ts
antomor: It's the private key corresponding to one of the preloaded accounts. You can see it [here](https://github.com/rsksmart/rif-relay/blob/0e9c2d506483096e8549ed125145720b7ec6a8ad/test/EnvelopingUtils.test.ts#L183) and it corresponds to account at index 9 of the accounts retrieved with eth_accounts from regtest. I think we should: 
1. store it in a constant since it's used elsewhere
2. add a comment to explicitly describe where it comes from.

PR #242, opened 2023-02-07
PP-634/registration-manager-tests

franciscotobar:
## What

- Migrate the integration tests to hardhat and ethers.js

## Why

- Project modernization

## Refs

- [PP-634](https://rsklabs.atlassian.net/browse/PP-634)


test/relayserver/RegistrationManager.test.ts
```diff
@@ -0,0 +1,724 @@
...
+
+        await evmMineMany(unstakeDelay);
+
+        const latestBlock = await provider.getBlock('latest');
+
+        const receipts = await relayServer._worker(latestBlock.number);
+
+        const gasPrice = await provider.getGasPrice();
+
+        // TODO: these two hard-coded indexes are dependent on the order of operations in 'withdrawAllFunds'
```
antomor, line 568: Which indexes are we referring to?

test/relayserver/RegistrationManager.test.ts
```diff
@@ -0,0 +1,724 @@
...
+        contracts: {
+          relayHubAddress: relayHub.address,
+        },
+      });
+      relayServer = await getInitiatedServer({ relayOwner });
+    });
+
+    afterEach(function () {
+      config.util.extendDeep(config, originalConfig);
+    });
```
antomor, line 711: why don't we move this afterEach to an upper level? If so, we wouldn't need to repeat it each time

antomor (approved):
@franciscotobar LGTM, just left some questions.

test/relayserver/RegistrationManager.test.ts
franciscotobar: I just keep it for structure, but there are some  afterEach that are not really necessary. I just have it like that to try to be more realistic.

test/relayserver/RegistrationManager.test.ts
franciscotobar: I removed the comment, before this had it, since they were sending the index[0] but its not longer necessary since each test its encapsulated.

test/relayserver/RegistrationManager.test.ts
franciscotobar: updated

test/relayserver/RegistrationManager.test.ts
franciscotobar: updated

antomor (approved):
LGTM

PR #245, opened 2023-02-28
PP-686: gas estimation task

ironFe93:
## What

Migrate relay hub gas estimation tests into a script/task.

## Why

It was not testing anything to begin with

**Note:
The reason why I went for a script and not a task is because the logic relies on some functionalities from test/utils/testUtils.ts  and unfortunately this file imports 'hardhat' in order to access ethers. Maybe importing ethers directly will allow it? I will attempt to convert into a task shortly.


ironFe93:
In order to make this script into a proper task, we would have to stop depending on testUtils (because test utils imports hardhat) or not use hardhat at all in testUtils (which would defeat half the purpose of this migration). Another option would be to extract out only the functions that are needed by this script. However this would cause code duplication and a few of these functions do rely on ethers.

scripts/RelayHubGasEstimation.ts
```diff
@@ -0,0 +1,641 @@
...
+  }) {
+    // gas estimation fit
+    // 5.10585×10^-14 x^3 - 2.21951×10^-8 x^2 + 1.06905 x + 92756.3
+    // 8.2808×10^-14 x^3 - 8.62083×10^-8 x^2 + 1.08734 x + 36959.6
+    const a0 = Number('35095.980');
+    const a1 = Number('1.098');
+    const estimatedCost =
+      a1 * (internalDestinationCallCost + internalTokenCallCost) + a0;
+    console.log('The predicted total cost is: ', estimatedCost);
+  }
```
antomor, line 186: I'd remove this function because:
1. this estimation is quite different from the real cost
2. the values used in the [client](https://github.com/rsksmart/rif-relay-client/blob/main/src/gasEstimator/utils.ts#L93) are different from those used here

scripts/RelayHubGasEstimation.ts
```diff
@@ -0,0 +1,641 @@
...
+    noCorrection: true,
+  });
+
+  const callWithoutRelay = await recipient.emitMessage(message);
+  const callWithoutRelayReceipt = await callWithoutRelay.wait();
+  const cumulativeGasUsedWithoutRelay =
+    callWithoutRelayReceipt.cumulativeGasUsed.toNumber();
+  const gasOverhead = cumulativeGasUsed - cumulativeGasUsedWithoutRelay;
+  console.log(
+    `Destination Call without enveloping - Gas Used: ${callWithoutRelayReceipt.gasUsed.toNumber()}, Cummulative Gas Used: ${cumulativeGasUsedWithoutRelay}`
```
antomor, line 453:
```suggestion
    `Destination Call without enveloping - Gas Used: ${callWithoutRelayReceipt.gasUsed.toNumber()}, Cumulative Gas Used: ${cumulativeGasUsedWithoutRelay}`
```

scripts/RelayHubGasEstimation.ts
```diff
@@ -0,0 +1,641 @@
...
+      gasPrice,
+    });
+
+  const nonceAfter = await forwarder.nonce();
+  expect(nonceBefore.add(1).toNumber()).to.equal(nonceAfter.toNumber());
+
+  const txReceipt = await relayCallResult.wait();
+
+  console.log(
+    `Gas Used: ${txReceipt.gasUsed.toString()} - Cummulative Gas Used: ${txReceipt.cumulativeGasUsed.toString()}\n`
```
antomor, line 518:
```suggestion
    `Gas Used: ${txReceipt.gasUsed.toString()} - Cumulative Gas Used: ${txReceipt.cumulativeGasUsed.toString()}\n`
```

scripts/RelayHubGasEstimation.ts
```diff
@@ -0,0 +1,641 @@
...
+        tokenContract,
+        gas: internalDestinationCallCost,
+        ...tokenGas,
+      },
+    });
+
+    return completeReq;
+  }
+
+  async function estimateGasOverhead(fees: string) {
```
antomor, line 585: I know the name is the same as before, but I'd rename it to
```suggestion
  async function estimateTokenTransferGasOverhead(fees: string) {
```

scripts/RelayHubGasEstimation.ts
```diff
@@ -0,0 +1,641 @@
...
+      smartWalletTemplate,
+      false,
+      owner,
+    )) as SmartWalletFactory
+  
+    forwarder = (await createSupportedSmartWallet({
+      relayHub: relayHubSigner.address,
+      factory,
+      owner,
+      sender: relayHubSigner,
```
antomor, line 113: Shouldn't the worker be the sender?

antomor (changes requested):
@ironFe93 Thank you for the changes; I've left some minor comments/questions. Other than them, I'd also include the estimation of a smart wallet deployment (we do already deploy it, but we don't log the gas spent for it)

scripts/RelayHubGasEstimation.ts
ironFe93: done

scripts/RelayHubGasEstimation.ts
ironFe93: I attempted this but it breaks. We would have to change the relayHub as well to be the worker.

antomor: yes, you're right. It looks like it fails because we're sending the request directly to the SmartWalletFactory. Although here we're deploying a SW, this is not the real process, because the call should be performed against `RelayHub.deployCall`. I'm going to create a task for this to cover this specific scenario.

PR #247, opened 2023-03-14
PP719: Adapt estimation script to calculate the cost of a deployment

AndresQuijano:
## What

- Modify RelayHubGasEstimation script to, additionally, print the cost related to deploying a smart wallet.

## Why

- Have a more complete tool to analyze gas consumption.

## Refs

- [PP-719](https://rsklabs.atlassian.net/browse/PP-719)


scripts/RelayHubGasEstimation.ts
```diff
@@ -54,374 +63,429 @@ function logTitle(title: string) {
...
-    );
+  return {
+    forwarder,
+    recipient,
+    relayHub,
+    owner,
+    relayRequest,
+    factory,
+    token,
+    verifier
```
antomor, line 151: if we return those objects here, why do we need to have them also defined globally? I'd remove the global definition to have them returned here only

scripts/RelayHubGasEstimation.ts
```diff
@@ -54,374 +63,429 @@ function logTitle(title: string) {
...
+const deployContract = <Contract>(contract: string) => {
+  return ethers
+    .getContractFactory(contract)
+    .then((contractFactory) => contractFactory.deploy() as Contract);
+};
 
-  async function printGasStatus(receipt: ContractReceipt) {
-    const [firstAccount] = (await ethers.getSigners()) as [SignerWithAddress]
-    await token.mint('1000', firstAccount.address)
+async function setupRelayHub(relayHub: RelayHub) {
```
antomor, line 77: Why do we define the functions with arrow functions and other as normal functions? I think we should try to be as much consistent as possible, so unless there's a specific reason, I'd apply always the same strategy to create functions.

scripts/RelayHubGasEstimation.ts
antomor, line 1: I'd rename this file to something more meaningful (e.g.: `GasEstimation.ts`)

scripts/RelayHubGasEstimation.ts
```diff
@@ -54,374 +63,429 @@ function logTitle(title: string) {
...
+    )
+  ).to.equal(true, 'Worker did not receive payment');
+};
 
-    return {
-      nonceBefore,
-      forwarderInitialBalance,
-      relayWorkerInitialBalance,
-    };
+const assertNonce = async (nonceBefore: BigNumber) => {
```
antomor, line 217:
```suggestion
const assertForwarderNonce = async (nonceBefore: BigNumber) => {
```

scripts/RelayHubGasEstimation.ts
```diff
@@ -505,122 +567,101 @@ const estimateGas = async () => {
...
+  await deployAndSetup();
+  await runGasEstimationScenarios();
+
+  logTitle('Deploy estimation without token payment');
+  await runDeployEstimation('0');
+  console.log('\nROUND 2');
+  await runDeployEstimation('0');
+
+  logTitle('Deploy estimation with token payment');
+  const TOKEN_AMOUNT = '100000000000000000000';
```
antomor, line 660: I think we could use [the `ethers.util.parseUnit` function](https://docs.ethers.org/v5/api/utils/display-logic/#utils-parseUnits) to make this number more readable and use the ether representation. 
```suggestion
  const TOKEN_AMOUNT = ethes.utils.parseUnit('100');
```

scripts/RelayHubGasEstimation.ts
```diff
@@ -54,374 +63,429 @@ function logTitle(title: string) {
...
 
   printGasAnalysis(relayCallProcessSecondResult);
-  
-  logTitle('Gas estimation tests for SmartWallet')
-  
-  message = 'RIF Enveloping '.repeat(32);
-  ({ recipient, forwarder, owner, relayRequest, relayHub } = await deployAndSetup());
+}
+
+async function runGasEstimation() {
```
antomor, line 486: I'd remove this at all. It's not clear to me why we have it.

scripts/RelayHubGasEstimation.ts
```diff
@@ -505,122 +567,101 @@ const estimateGas = async () => {
...
-    const estimatedTokenPaymentGas = await ethers.provider.estimateGas({
-      from: forwarder.address,
-      to: token.address,
-      data: token.interface.encodeFunctionData('transfer', [
-        relayWorker.address,
-        fees,
-      ]),
-    });
+async function runDeployEstimation(tokenAmount = '0') {
+  const {owner, relayHub,factory, token, verifier} = await deployAndSetup();
```
antomor, line 578: This function also deploys the SW in without using the relayHub, and it's totally unnecessary in this case. I'd change that function to avoid it (also because the log created is quite confusing IMHO)

scripts/RelayHubGasEstimation.ts
```diff
@@ -505,122 +567,101 @@ const estimateGas = async () => {
...
 
-    // forge the request
-    const completeReq: RelayRequest = await forgeRequest(
-      transferReceiver.address,
-      balanceToTransfer,
-      fees,
-      token, relayRequest,
-      forwarder
-    );
+  console.log('Total gas used on deploy: ', gasUsed.toString());
```
antomor, line 622: We're printing the total gas used for the deployment without highlighting the information (I'd use different colors compared to the ones used for the transfer overhead)

antomor (changes requested):
@AndresQuijano Thank you for the changes. I think we should simplify it a bit. Furthermore, I was wondering if we should create different tasks for each scenario that we want to cover. I've realised that there are conflicts and that this file isn't covered when running the lint-staged tests. I'm going to update this PR to run the lint-staged checks on it and to resolve the conflicts also.

scripts/RelayHubGasEstimation.ts
AndresQuijano: Done.

scripts/RelayHubGasEstimation.ts
AndresQuijano: Done.

scripts/RelayHubGasEstimation.ts
AndresQuijano: Done.

scripts/RelayHubGasEstimation.ts
AndresQuijano: Done.

scripts/RelayHubGasEstimation.ts
```diff
@@ -445,27 +509,25 @@ const estimateGas = async () => {
...
+
   console.log(
     `Destination Call with enveloping - Gas Used: ${gasUsedRound2.toNumber()}, CumulativeGasUsed: ${cumulativeGasUsedRound2}\n`
   );
+}
 
-  logTitle('Gas estimation scenarios');
-
-  await deployAndSetup();
+async function runGasEstimationScenarios() {
```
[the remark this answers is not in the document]
AndresQuijano: I changed the name of the function for estimateRelayCost(). Let me know if you think the name is fine.

scripts/RelayHubGasEstimation.ts
AndresQuijano: Removed.

scripts/RelayHubGasEstimation.ts
AndresQuijano: Refactored.

scripts/RelayHubGasEstimation.ts
AndresQuijano: I'm not really fan of using colors in log. I simplified the output of the script without colors. Have a look at it and give your opinion.

scripts/RelayHubGasEstimation.ts
```diff
@@ -505,122 +567,101 @@ const estimateGas = async () => {
...
+  logTitle('Deploy estimation without token payment');
+  await runDeployEstimation('0');
+  console.log('\nROUND 2');
+  await runDeployEstimation('0');
+
+  logTitle('Deploy estimation with token payment');
+  const TOKEN_AMOUNT = '100000000000000000000';
+  await runDeployEstimation(TOKEN_AMOUNT);
+  console.log('\nROUND 2');
+  await runDeployEstimation(TOKEN_AMOUNT);
```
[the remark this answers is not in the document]
AndresQuijano: Done.

antomor (changes requested):
@AndresQuijano It looks much clearer the script now, thank you. I've just left a comment because I realised that the transfer cost changes if we use the same account as sender and receiver of a transfer.

scripts/GasEstimation.ts
```diff
@@ -0,0 +1,435 @@
...
+  return {
+    relayHub,
+    owner,
+    smartWalletFactory,
+    token,
+    verifier,
+  };
+}
+
+async function getTransferEstimationWithoutRelay(token: UtilToken) {
```
franciscotobar, line 109: I don't understands the purpose of this function, or maybe we can improve the name.

scripts/GasEstimation.ts
```diff
@@ -0,0 +1,435 @@
...
+  const gasCorrectionFactor = BigNumber.from(ESTIMATED_GAS_CORRECTION_FACTOR);
+
+  return gasCorrectionFactor.mul(
+    estimation.gt(internalCorrection)
+      ? estimation.sub(internalCorrection)
+      : estimation
+  );
+}
+
+async function getCurrentBalances(
```
franciscotobar, line 177: Im not pretty sure what is the benefit of this function.

scripts/GasEstimation.ts
```diff
@@ -0,0 +1,435 @@
...
+    factory: smartWalletFactory,
+    owner,
+    sender: relayHubSigner,
+  })) as SmartWallet;
+
+  const baseRelayRequest = {
+    request: {
+      relayHub: relayHub.address,
+      to: recipient.address,
+      data: '0xdeadbeef',
```
franciscotobar, line 276: what is this data?

scripts/GasEstimation.ts
```diff
@@ -0,0 +1,435 @@
...
+      tokenAmount: fees,
+      tokenContract,
+      gas: estimatedDestinationCallGasCorrected.toString(),
+      tokenGas,
+    },
+  });
+}
+
+async function estimateRelayCost(fees = '0') {
+  const recipient = await deployContract<TestRecipient>('TestRecipient');
```
franciscotobar, line 261: Can this be parametized?

scripts/GasEstimation.ts
```diff
@@ -0,0 +1,435 @@
...
+      ...overrideRelayRequest.relayData,
+    },
+  };
+}
+
+function logTitle(title: string) {
+  console.log(`${UNDERLINE}\n${title.toUpperCase()}${RESET}`);
+}
+
+async function deployContract<Contract>(contract: string) {
```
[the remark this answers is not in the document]
AndresQuijano: Thanks, changed.

scripts/GasEstimation.ts
```diff
@@ -0,0 +1,435 @@
...
+
+async function setupRelayHub(relayHub: RelayHub) {
+  await relayHub
+    .connect(relayOwner)
+    .stakeForAddress(relayManager.address, 1000, {
+      value: ethers.utils.parseEther('2'),
+    });
+
+  await relayHub.connect(relayManager).addRelayWorkers([relayWorker.address]);
+  await relayHub.connect(relayManager).registerRelayServer('http://relay.com');
```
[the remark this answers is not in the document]
AndresQuijano: Done.

scripts/GasEstimation.ts
AndresQuijano: Yes, makes no sense to have this on a function. I just deleted it.

scripts/GasEstimation.ts
AndresQuijano: Just dummy data. Dead beef, probably some meat lover wrote it.

scripts/GasEstimation.ts
```diff
@@ -0,0 +1,435 @@
...
+    },
+    relayData: {
+      gasPrice: GAS_PRICE,
+      feesReceiver: relayWorker.address,
+      callForwarder: smartWallet.address,
+      callVerifier: verifier.address,
+    },
+  };
+
+  await token.mint(TOKEN_AMOUNT_TO_TRANSFER + '00', smartWallet.address);
```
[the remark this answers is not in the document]
AndresQuijano: Because the smart wallet needs the value to be transferred and  extra balance  to pay for the transaction.

scripts/GasEstimation.ts
AndresQuijano: It can but will add not value. The string used here is very descriptive and is not being reused.

scripts/GasEstimation.ts
```diff
@@ -0,0 +1,435 @@
...
+  const penalizer = await deployContract<Penalizer>('Penalizer');
+  const verifier = await deployContract<TestVerifierEverythingAccepted>(
+    'TestVerifierEverythingAccepted'
+  );
+  const token = await deployContract<UtilToken>('UtilToken');
+  const smartWalletTemplate = await deployContract<SmartWallet>('SmartWallet');
+  const relayHub = await deployRelayHub(penalizer.address);
+  await setupRelayHub(relayHub);
+
+  await fundedAccount.sendTransaction({
```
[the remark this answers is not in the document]
AndresQuijano: Initially all counts where global and I changed that. I just let the ones obtained from ethers.getAccounts for ease to use. We can talk about this.

scripts/GasEstimation.ts
AndresQuijano: I removed the recipient and the data related to the call.

franciscotobar (approved):
LGTM

PR #250, opened 2023-05-04
fix/verify_with_relay_hub

franciscotobar:
## What

- Update the static client validation with the relayHub

## Why

- The server tries increasing the gas limit with an increase factor and the client does not. 

## Refs

- https://github.com/rsksmart/rif-relay-client/pull/86
- https://github.com/rsksmart/rif-relay-server/pull/107


package.json
```diff
@@ -88,9 +88,9 @@
     "@metamask/eth-sig-util": "^5.0.2",
     "@nomicfoundation/hardhat-toolbox": "^2.0.0",
     "@nomiclabs/hardhat-ethers": "^2.2.1",
-    "@rsksmart/rif-relay-client": "github:rsksmart/rif-relay-client",
+    "@rsksmart/rif-relay-client": "github:rsksmart/rif-relay-client#fix/verify_with_relay_hub",
     "@rsksmart/rif-relay-contracts": "github:rsksmart/rif-relay-contracts",
-    "@rsksmart/rif-relay-server": "github:rsksmart/rif-relay-server",
+    "@rsksmart/rif-relay-server": "github:rsksmart/rif-relay-server#fix/verify_with_relay_hub",
```
antomor, line 93: We need to update it after merging the PRs

test/relayserver/RelayServer.test.ts
```diff
@@ -484,58 +481,6 @@ describe('RelayServer', function () {
       });
     });
 
-    describe('maxPossibleGasWithViewCall', function () {
```
antomor, line 487: Why did we remove this test?

antomor (approved):

test/relayserver/RelayServer.test.ts
franciscotobar: this function does not longer exists in the server, I can move it to the server, since I dont think that removing it is the best choice.

franciscotobar: updated, I return the function to avoid breaking changes.

antomor (approved):
LGTM

package.json
franciscotobar: updated

PR #260, opened 2024-01-29
PP-885/explore-solution-b2

franciscotobar:
## What

- Implements solution [B2](https://docs.google.com/document/d/10N8jV41KwOXdoiCcSGHRvS5TYGyA_1KkVpee75eG3Gg/edit#heading=h.ddb8zsm0g2vy)

## Why

- Necessary to execute a contract while deploying a smart wallet

## Refs

-  [PP-885](https://rsklabs.atlassian.net/browse/PP-885)
- https://github.com/rsksmart/rif-relay-contracts/pull/149
- https://github.com/rsksmart/rif-relay-client/pull/90
- https://github.com/rsksmart/rif-relay-server/pull/140


.husky/pre-push
```diff
@@ -1,4 +1,4 @@
 #!/usr/bin/env sh
 . "$(dirname -- "$0")/_/husky.sh"
 
-npm run test
+npm run test -- --network regtest
```
antomor, line 4: ❤️

antomor (changes requested):

package.json
```diff
@@ -88,9 +88,10 @@
     "@metamask/eth-sig-util": "^5.0.2",
     "@nomicfoundation/hardhat-toolbox": "^2.0.0",
     "@nomiclabs/hardhat-ethers": "^2.2.1",
-    "@rsksmart/rif-relay-client": "github:rsksmart/rif-relay-client",
-    "@rsksmart/rif-relay-contracts": "github:rsksmart/rif-relay-contracts",
-    "@rsksmart/rif-relay-server": "github:rsksmart/rif-relay-server",
+    "@openzeppelin/contracts": "^3.4.0",
```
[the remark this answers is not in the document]
franciscotobar: yes we need it, since its use by the local contracts, if you try the main branch you will see that it will fail without it.

test/relayclient/RelayClient.test.ts
```diff
@@ -555,6 +561,73 @@ describe('RelayClient', function () {
         });
         expect(to).to.be.equal(relayHub.address);
       });
+
+      describe('with contract execution during deploy', function () {
```
antomor, line 565: We don't need to specify the deployment since we're already in the deployment describe section
```suggestion
      describe('with contract execution', function () {
```

test/utils/TestUtils.ts
```diff
@@ -167,52 +198,51 @@ const deployRelayHub = async (
...
     relayVerifier,
   };
 };
 
-const createSmartWalletFactory = async (
-  template: IForwarder,
-  isCustom = false,
+const createSmartWalletFactory = async <T extends SupportedSmartWalletFactory>(
+  template: SupportedSmartWallet | MinimalBoltzSmartWallet,
+  type: SupportedType = 'Default',
```
antomor, line 233: If we declare `type` as the latest argument, we could avoid specifying it when it's value is `Default`, otherwise it doesn't make sense to use a default value here

test/relayserver/RelayServer.test.ts
```diff
@@ -168,15 +177,12 @@ describe('RelayServer', function () {
...
-        const userDefinedRelayRequest = createUserDefinedRequest(
-          IS_DEPLOY_REQUEST,
-          {
-            from: owner.address,
-            to: recipient.address,
-            data: encodedData,
-            nonce: 1,
-          }
-        );
+        const userDefinedRelayRequest = createUserDefinedRequest(false, {
```
antomor, line 180: I know you didn't introduce this change, but what do you think if we create 2 functions `createDeployUserDefinedRequest` and `createRelayUserDefinedRequest` instead of using the flag?

test/relayserver/RelayServer.test.ts
```diff
@@ -429,20 +417,79 @@ describe('RelayServer', function () {
...
+
+    async function prepareVerifier<
+      F extends SupportedSmartWalletFactory,
+      V extends SupportedDeployVerifier
+    >(
+      fundedAccount: SignerWithAddress,
+      type: SupportedType
+    ): Promise<{ factory: F; verifier: V }> {
+      const smartWalletTemplate: SupportedSmartWallet = await deployContract(
+        `${type === 'Default' ? '' : type}SmartWallet`
```
antomor, line 434: We use this in other places. Couldn't we create a function for this?
e.g.:
```
const getSmartWalletTemplate = (type) => `${type === 'Default' ? '' : type}SmartWallet`;
```

test/smartwallet/baseSmartWallet.test.ts
```diff
@@ -44,12 +40,8 @@ const INITIAL_SMART_WALLET_RBTC_AMOUNT = 50;
...
 const RBTC_AMOUNT_TO_TRANSFER = hardhat.utils.parseEther('1');
 
-const CUSTOM_SMART_WALLET_TYPE: TypeOfWallet = 'CustomSmartWallet';
-const SMART_WALLET_TYPE: TypeOfWallet = 'SmartWallet';
-const TYPES_OF_WALLETS: TypeOfWallet[] = [
-  CUSTOM_SMART_WALLET_TYPE,
-  SMART_WALLET_TYPE,
-];
+type TypeOfWallet = 'Custom' | 'Default';
+const TYPES_OF_WALLETS: TypeOfWallet[] = ['Default', 'Custom'];
```
antomor, line 44: Shouldn't we add here also `BoltzSmartWallet` at least?

test/smartwallet/baseSmartWallet.test.ts
```diff
@@ -44,12 +40,8 @@ const INITIAL_SMART_WALLET_RBTC_AMOUNT = 50;
 const TOKEN_AMOUNT_TO_TRANSFER = 1;
 const RBTC_AMOUNT_TO_TRANSFER = hardhat.utils.parseEther('1');
 
-const CUSTOM_SMART_WALLET_TYPE: TypeOfWallet = 'CustomSmartWallet';
-const SMART_WALLET_TYPE: TypeOfWallet = 'SmartWallet';
-const TYPES_OF_WALLETS: TypeOfWallet[] = [
-  CUSTOM_SMART_WALLET_TYPE,
-  SMART_WALLET_TYPE,
-];
+type TypeOfWallet = 'Custom' | 'Default';
```
antomor, line 43: I think we defined the SmartWallet types elsewhere

test/RelayHub.test.ts
```diff
@@ -976,6 +982,377 @@ describe('RelayHub', function () {
...
+          nextWalletIndex++;
+
+          const deployCall = relayHub
+            .connect(relayWorker)
+            .deployCall(deployRequest, signature, { gasLimit });
+
+          await expect(deployCall).not.to.be.rejected;
+        });
+
+        it.skip('should fail if not enough gas to pay for native transfer', async function () {
```
antomor, line 1325: why are we skipping this test?

antomor: Is this related to the automatic token gas set by the VM? If so, I'd either change the test not to expect a fail and adding a comment to describe why we don't expect this to fail

test/relayserver/RelayServer.test.ts
```diff
@@ -666,6 +699,199 @@ describe('RelayServer', function () {
...
+
+          await expect(
+            relayServer.createRelayTransaction(
+              stringifyEnvelopingTx(envelopingTxRequest)
+            )
+          ).to.be.rejectedWith('Claiming value lower than fees');
+        });
+
+        // TODO - Should bubble up error but its failing
+        it.skip('should fail if destination contract throws error', async function () {
```
antomor, line 785: In this case, I think that the good part is that still the promise rejects as expected, but we don't get the proper error message. I'd try to investigate a bit more on the reason, otherwise, I wouldn't skip the test, but I'd add a comment on the expected error clarifying it.

contracts/BoltzDeployVerifier.sol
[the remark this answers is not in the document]
franciscotobar: updated

test/relayclient/RelayClient.test.ts
franciscotobar: updated

test/relayserver/RelayServer.test.ts
franciscotobar: updated

test/relayserver/RelayServer.test.ts
franciscotobar: not really, but I updated other parts to reuse an existing function.

test/smartwallet/baseSmartWallet.test.ts
franciscotobar: updated

test/smartwallet/baseSmartWallet.test.ts
franciscotobar: updated

test/RelayHub.test.ts
franciscotobar: it is related to that, let me fix it.

test/relayserver/RelayServer.test.ts
franciscotobar: correct it fails but with a `transaction reverted`, let me fix it then. 

We already know the issue and how to fix it, we discuss fixing in a different task.

test/RelayHub.test.ts
franciscotobar: updated

test/relayserver/RelayServer.test.ts
franciscotobar: updated

test/RelayHub.test.ts
```diff
@@ -1322,7 +1319,10 @@ describe('RelayHub', function () {
           await expect(deployCall).not.to.be.rejected;
         });
 
-        it.skip('should fail if not enough gas to pay for native transfer', async function () {
+        // When the tokenGas is set to 0, we should expect to fail but the VM provide a minimum gas of 2300 by default
+        // that covers the transfer cost, this will only fail if the destination, its a contract that consumes more than
```
antomor, line 1323:
```suggestion
        // that covers the transfer cost, this will only fail if the destination is a contract that consumes more than
```

test/utils/TestUtils.ts
```diff
@@ -628,3 +641,5 @@ export type {
   SupportedType,
   SupportedDeployVerifier,
 };
+
+// delete prepareRelayTransaction
```
antomor, line 645: @franciscotobar Is this comment still valid?

test/RelayHub.test.ts
franciscotobar: updated

test/utils/TestUtils.ts
franciscotobar: Updated

antomor (approved):
LGTM

PR #298, opened 2026-01-29
Refactoring devportal-update.yml  to use GHApp

scguaquetam:
Refactoring devportal-update.yml automation script for using GHApp instead of PAT

## What

- Update dev portal action

## Why

- To use GH apps instead of PAT into the flow for update Developer Portal content when modifying Readme


ezequiel-rodriguez (approved):
LGTM

